PR #17, opened 2019-10-08
Make impl_params a attribute of the predefined operators

pmeier:
This makes `impl_params` direct or indirect an attribute for the following classes

- `GatysEckerBethge2015ContentLoss`
- `GatysEckerBethge2015StyleLoss`
- `GatysEckerBethge2015NST`
- `GatysEtAl2017StyleLoss`
- `GatysEtAl2017NSTPyramid`
- `GatysEtAl2017SpatialControlNSTPyramid`
- `LiWand2016ContentLoss`
- `LiWand2016StyleLoss`
- `LiWand2016Regularizer`
- `LiWand2016NSTPyramid`

Additionally, the `build_levels()` methods of the `*Pyramid`s now does not take an input argument, but instead relies on the new attribute. Furthermore,  `build_levels()` method is now called in the constructor, eliminating the need to call it manually, since all parameters are fixed anyway.

### DIFF
--- a/pystiche/papers/gatys_et_al_2017.py
+++ b/pystiche/papers/gatys_et_al_2017.py
@@ -62,7 +62,11 @@ class GatysEtAl2017StyleLoss(GramEncodingComparisonOperator):
             parameters given in the paper are used.
     """
 
-    def __init__(self, encoder, impl_params=True):
+    def __init__(self, encoder=None, impl_params=True):
+        if encoder is None:
+            encoder = get_encoder()
+        self.impl_params = impl_params
+
         name = "Style loss (Gram)"
         layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1", "relu_5_1")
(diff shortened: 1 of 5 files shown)


pystiche/papers/gatys_et_al_2017.py
```diff
@@ -174,22 +178,19 @@ def __init__(self, num_guides, impl_params=True, guide_names=None):
...
 class _GatysEtAl2017NSTPyramidBase(ImageOptimizerPyramid):
-    def __init__(self, nst):
+    def __init__(self, nst, impl_params):
         super().__init__(nst)
         self.nst = nst
+        self.impl_params = impl_params
+        self.build_levels()
 
-    def build_levels(self, impl_params=True):
+    def build_levels(self):
```
pmeier: The removal of this input parameter breaks BC. Same for `LiWand2016NSTPyramid`.

PR #32, opened 2020-01-09
fixed pillow import

pmeier:
Corrects #29 

Change the required `pillow` version from `<= 6.2.0` to `< 7.0.0` to include the most recent versions in between.

### DIFF
--- a/pystiche/image/io.py
+++ b/pystiche/image/io.py
@@ -36,6 +36,6 @@ def write_image(
 
 
 def show_image(
-    image: torch.Tensor, mode: Optional[str] = None, title: Optional[str] = None,
+    image: torch.Tensor, mode: Optional[str] = None, title: Optional[str] = None
 ):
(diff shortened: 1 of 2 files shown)


pystiche/image/io.py
```diff
@@ -36,6 +36,6 @@ def write_image(
 
 
 def show_image(
-    image: torch.Tensor, mode: Optional[str] = None, title: Optional[str] = None,
+    image: torch.Tensor, mode: Optional[str] = None, title: Optional[str] = None
```
pmeier: Unrelated change from `black`

hugovk:
[torchvision v0.5.0](https://github.com/pytorch/vision/releases/tag/v0.5.0) has been released with the fix:

1. Require `torchvision>=0.5.0`
2. If Pillow was temporarily pinned, remove the pin

PR #47, opened 2020-01-31
Added special guidance for MRFOperator

pmeier:
Due to the guiding large areas of the images are zero and thus many patches carry no information. Thus, these patches can be removed from the target and input representation reducing the computing cost and memory during loss calculation.


pmeier:
CI build since checks are ATM not displayed in the GitHub UI:

https://travis-ci.org/pmeier/pystiche/builds/644341269

PR #88, opened 2020-03-03
Add paper Texture Networks: Feed-forward Synthesis of Textures and Stylized Images

jbueltemeier:
Texture Networks: Feed-forward Synthesis of Textures and Stylized Images
Dmitry Ulyanov, Vadim Lebedev, Andrea Vedaldi, Victor Lempitsky
https://github.com/DmitryUlyanov/texture_nets


pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+__all__ = [
+    "ulyanov_et_al_2016_transformer_optim_loop",
+    "ulyanov_et_al_2016_transformer",
+    "ulyanov_et_al_2016_perceptual_loss",
+    "ulyanov_et_al_2016_dataset",
+    "ulyanov_et_al_2016_image_loader",
+    "ulyanov_et_al_2016_training",
+    "ulyanov_et_al_2016_images",
+    "ulyanov_et_al_2016_stylization",
+]
```
pmeier: Only top level functions, i.e. `transformer`, `loss`, `images`, `dataset`, `training` and `stylization`, should be placed here.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+        else:
+            raise NotImplementedError
+
+    return transformer
+
+
+def ulyanov_et_al_2016_training(
+    content_image_loader: DataLoader,
+    style: Union[str, torch.Tensor],
+    impl_mode: str = "paper",
```
pmeier: We should stick to `impl_params: bool` and `instance_norm: bool`. What exactly both parameters mean should be part of the documentation.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+    instance_norm: bool = True,
+    mode: str = "texture",
```
pmeier: Please change this to `stylization: bool`, which defaults to `True`.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+    "ulyanov_et_al_2016_perceptual_loss",
+    "ulyanov_et_al_2016_dataset",
+    "ulyanov_et_al_2016_image_loader",
+    "ulyanov_et_al_2016_training",
+    "ulyanov_et_al_2016_images",
+    "ulyanov_et_al_2016_stylization",
+]
+
+
+def ulyanov_et_al_2016_transformer_optim_loop(
```
pmeier: If I'm not mistaken the whole purpose of this custom `optim_loop` is change of the learning rate (`lr`) at the end. Is that assessment correct? If yes, we should create an `epoch_optim_loop` and handle this with a `lr_scheduler`.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def ulyanov_et_al_2016_stylization(
+    input_image: torch.Tensor,
+    transformer: Union[nn.Module, str],
+    impl_mode: str = "paper",
+    instance_norm: bool = None,
+    weights: str = "pystiche",
+    sample_size: int = 256,
+    postprocessor: ulyanov_et_al_2016_postprocessor = ulyanov_et_al_2016_postprocessor(),
```
pmeier: Mutable default argument!

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+
+    style_transform = ulyanov_et_al_2016_style_transform(impl_mode=impl_mode)
+    style_transform = style_transform.to(device)
+    style_image = style_transform(style_image)
+    style_image = batch_up_image(style_image, loader=content_image_loader)
+
+    criterion.set_style_image(style_image)
+
+    def criterion_update_fn(
+        input_image, criterion, preprocessor=ulyanov_et_al_2016_preprocessor()
```
pmeier: Never use mutable objects as default argument! See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,293 @@
...
+) -> FiniteCycleBatchSampler:
+    if num_batches is None:
+        if impl_mode == "texturenetsv1":
+            num_batches = 3000 if mode == "style" else 1500
+        elif impl_mode == "paper":
+            num_batches = 2000
+        elif impl_mode == "master":
+            num_batches = 50000
+        else:
+            raise NotImplementedError
```
pmeier: A [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError) implies that this method is abstract, i.e. should be overwritten in a derived class. Thus, a [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError) would be a better fit here.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,293 @@
...
+
+    if batch_size is None:
+        if impl_mode == "texturenetsv1":
+            batch_size = 4 if mode == "style" else 16
+        elif impl_mode == "paper":
+            batch_size = 16
+        elif impl_mode == "master":
+            batch_size = 1
+        else:
+            raise NotImplementedError
```
pmeier: `NotImplementedError` vs. `RuntimeError`!

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+from pystiche.ops import (
+    MSEEncodingOperator,
+    GramOperator,
+    MultiLayerEncodingOperator,
+    TotalVariationOperator,
+)
+from pystiche.loss import MultiOperatorLoss
+from .utils import ulyanov_et_al_2016_multi_layer_encoder
+
+
```
pmeier: The important functions, i.e. the functions imported by `core.py` should be included in `__all__`.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+):
+    if score_weight is None:
+        if impl_mode == "texturenetsv1":
+            score_weight = 6e-1
+        elif impl_mode == "paper":
+            score_weight = 1e0  # FIXME paper only alpha
+        elif impl_mode == "master":
+            score_weight = 1e0
+        else:
+            raise NotImplementedError
```
pmeier: `NotImplementedError` vs. `RuntimeError`!

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+    impl_mode: str = "paper",
+    multi_layer_encoder: Optional[MultiLayerEncoder] = None,
+    layer: str = "relu_4_2",
+    score_weight=None,
+):
+    if score_weight is None:
+        if impl_mode == "texturenetsv1":
+            score_weight = 6e-1
+        elif impl_mode == "paper":
+            score_weight = 1e0  # FIXME paper only alpha
```
pmeier: Please address this `FIXME`.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+):
+    if score_weight is None:
+        if impl_mode == "texturenetsv1":
+            score_weight = 1e3 if mode == "style" else 1e0
+        elif impl_mode == "paper":
+            score_weight = 1e0
+        elif impl_mode == "master":
+            score_weight = 1e0
+        else:
+            raise NotImplementedError
```
pmeier: `NotImplementedError` vs. `RuntimeError`!

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+    score_weight: float = None,
+    **total_variation_op_kwargs: Any,
+):
+    if score_weight is None:
+        if impl_mode == "texturenetsv1":
+            return None
+        elif impl_mode == "paper":
+            return None
+        elif impl_mode == "master":
+            score_weight = 0  # FIXME: tvLoss only in master but weight 0
```
pmeier: Does that mean that the TV loss is not used at all? If that is the case please remove it completely.

pystiche/papers/ulyanov_et_al_2016/utils.py
```diff
@@ -0,0 +1,32 @@
...
+) -> Optimizer:
+
+    if impl_mode == "texturenetsv1":
+        lr = 1e-1
+    elif impl_mode == "paper":
+        lr = 1e-1
+    elif impl_mode == "master":
+        lr = 1e-3
+    else:
+        raise NotImplementedError
```
pmeier: `NotImplementedError` vs. `RuntimeError`!

pystiche/papers/ulyanov_et_al_2016/core.py
jbueltemeier: In addition to the `lr`, the loss is also changed. I have already moved this to the criterion and create an `epoch_optim_loop`. However, I create an `optimizer` there, which I cannot pass to the `default_transformer_optim_loop`, because `get_optimizer` must be callable there. Would it make sense to change this so that an `optimizer` can also be transferred?

pystiche/papers/ulyanov_et_al_2016/loss.py
jbueltemeier: The paper does not provide any information on the learning rate (`lr`). This parameter is only given with a variable alpha without defined value.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,205 @@
...
+    "ulyanov_et_al_2016_perceptual_loss",
+    "ulyanov_et_al_2016_dataset",
+    "ulyanov_et_al_2016_image_loader",
+    "ulyanov_et_al_2016_training",
+    "ulyanov_et_al_2016_images",
+    "ulyanov_et_al_2016_stylization",
+]
+
+
+def ulyanov_et_al_2016_transformer_epoch_optim_loop(
```
pmeier: We should implement a `default_transformer_epoch_optim_loop()` and just parametrize it here if necessary. I will take that on and get back to you if I'm finished.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,297 @@
...
+    instance_norm: bool = True,
+    edge_size: Optional[int] = None,
+) -> ComposedTransform:
+    if edge_size is None:
+        edge_size = 256
+
+    if impl_params:
+        interpolation_mode = "bicubic" if instance_norm else "bilinear"
+    else:
+        interpolation_mode = "bilinear"  # FIXME: paper?
```
pmeier: Please address this `FIXME`.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,297 @@
...
+            is_grayscale = extract_num_channels(input_image) == 1
+            if is_grayscale:
+                return grayscale_to_fakegrayscale(input_image)
+            else:
+                return input_image
+
+    transforms = []
+    if impl_params:
+        if instance_norm:
+            # FIXME: RandomCrop here
```
pmeier: Please address this `FIXME`.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,297 @@
...
+                Resize((edge_size, edge_size), interpolation_mode="bicubic")
+            )
+        else:
+            transforms.append(
+                Resize((edge_size, edge_size), interpolation_mode="bilinear")
+            )
+    else:
+        transforms.append(
+            Resize((edge_size, edge_size), interpolation_mode="bilinear")
+        )  # FIXME: paper?
```
pmeier: Please address this `FIXME`.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,297 @@
...
+        ),
+        "kitty": DownloadableImage(
+            urljoin(readme_ulyanov, "kitty.jpg"),
+            md5="98262bd8f5ae25f8329158d2c2c66ad0",
+        ),
+    }
+
+    texture_base_ulyanov = urljoin(base_ulyanov, "textures/")
+    base_ulyanov_suppl_texture = "https://raw.githubusercontent.com/DmitryUlyanov/texture_nets/texture_nets_v1/supplementary//texture_models/"
+    # TODO: "https://www.cns.nyu.edu/~eero/texture/index.php#examples" license
```
pmeier: Please address this `TODO`.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,226 @@
...
+    ulyanov_et_al_2016_images,
+)
+from .utils import (
+    ulyanov_et_al_2016_preprocessor,
+    ulyanov_et_al_2016_postprocessor,
+    ulyanov_et_al_2016_optimizer,
+)
+
+__all__ = [
+    "ulyanov_et_al_2016_transformer_optim_loop",
```
pmeier: This should not be part of the outer interface.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,293 @@
...
+            else:
+                return input_image
+
+    transforms = []
+    if impl_mode == "texturenetsv1":
+        transforms.append(Resize((edge_size, edge_size), interpolation_mode="bilinear"))
+    elif impl_mode == "paper":
+        transforms.append(
+            Resize((edge_size, edge_size), interpolation_mode="bilinear")
+        )  # FIXME: paper?
```
pmeier: Please address this `FIXME`.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,293 @@
...
+    elif impl_mode == "master":
+        # FIXME: RandomCrop here
```
pmeier: Please address this `FIXME`.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,293 @@
...
+        ),
+        "kitty": DownloadableImage(
+            urljoin(readme_ulyanov, "kitty.jpg"),
+            md5="98262bd8f5ae25f8329158d2c2c66ad0",
+        ),
+    }
+
+    texture_base_ulyanov = urljoin(base_ulyanov, "textures/")
+    base_ulyanov_suppl_texture = "https://raw.githubusercontent.com/DmitryUlyanov/texture_nets/texture_nets_v1/supplementary//texture_models/"
+    # TODO: "https://www.cns.nyu.edu/~eero/texture/index.php#examples" license
```
pmeier: Please address this `TODO`.

pystiche/papers/ulyanov_et_al_2016/loss.py
pmeier: What has the learning rate `lr` to do with the content `score_weight`? If no information is given in the paper we should set it to `1e0` just as you did.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+
+    if layers is None:
+        if impl_mode == "texturenetsv1":
+            layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1", "relu_5_1")
+        elif impl_mode == "paper":
+            layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1", "relu_5_1")
+        elif impl_mode == "master":
+            layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1")
+        else:
+            raise NotImplementedError
```
pmeier: `NotImplementedError` vs. `RuntimeError`!

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,182 @@
...
+):
+    if score_weight is None:
+        if impl_mode == "texturenetsv1":
+            return None
+        elif impl_mode == "paper":
+            return None
+        elif impl_mode == "master":
+            score_weight = 0  # FIXME: tvLoss only in master but weight 0
+        else:
+            raise NotImplementedError
```
pmeier: `NotImplementedError` vs. `RuntimeError`!

pystiche/papers/ulyanov_et_al_2016/core.py
pmeier: See [my comment below](https://github.com/pmeier/pystiche/pull/88#discussion_r391459374)

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: The author of the paper has used RandomCrop (master branch) or has not given any information on pre-processing in his paper. Should therefore a scaling take place before the RandomCrop in this implementation?

pmeier: If no information is given in the paper we should always assume the simplest case. In this case that means that no preprocesing is perfomerd at all.

jbueltemeier: This will not be possible here. Due to the structure of the generator network and the associated downscaling and upscaling of the content image, the image should have a size which is a multiple of 2. For example 256x256 or 512x512. For this reason, the generated images in the paper also have the size of 256x256.

pmeier: If the images used in the paper are `256x256` why do we need cropping at all?

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,273 @@
...
+        edge_size = 256
+
+    if impl_params:
+        interpolation_mode = "bicubic" if instance_norm else "bilinear"
+    else:
+        interpolation_mode = "bilinear"
+
+    transforms = [
+        Resize(edge_size, edge="long", interpolation_mode=interpolation_mode),
+        CaffePreprocessing(),
```
pmeier: Shouldn't this be handled by a `preprocessor` in the high level function? With included `CaffePreprocessing`, the `style_transform` now only is applicable to an encoder with weights from `caffe`.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,273 @@
...
+    }
+    return DownloadableImageCollection(
+        {**texture_images, **content_images, **style_images},
+        root=root,
+        download=download,
+        overwrite=overwrite,
+    )
+
+
+def ulyanov_et_al_2016_dataset(
```
pmeier: What dataset are we using here? AFAIK no dataset was mentioned in the paper, correct? What dataset is the author using in his implementation?

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+
+
+def ulyanov_et_al_2016_image_loader(
+    dataset: Dataset,
+    impl_params: bool = True,
+    instance_norm: bool = True,
+    stylization: bool = True,
+    batch_sampler: Optional[Sampler] = None,
+    num_workers: int = 4,
```
pmeier: What is the default value for `num_workers` for `DataLoader`? We should also use this here. Using `4` seems arbitrary.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if layers is None:
+        if impl_params:
+            layers = (
+                ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1")
+                if instance_norm
+                else ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1", "relu_5_1")
+            )
+        else:
+            layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1", "relu_5_1")
```
pmeier: You could make this more readable by splitting just into two cases:
```python
if impl_params and not instance_norm:
    layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1")
else:
    layers = ("relu_1_1", "relu_2_1", "relu_3_1", "relu_4_1", "relu_5_1")
```

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,167 @@
...
+    def __init__(
+        self,
+        style_loss: MultiLayerEncodingOperator,
+        content_loss: MSEEncodingOperator = None,
+        stylization: bool = True,
+    ) -> None:
+
+        dict = OrderedDict([("style_loss", style_loss)])
+        if stylization:
+            dict["content_loss"] = content_loss
```
pmeier: There is no need to pass `stylization` here. You could simply check `if content_loss is not None`.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,167 @@
...
+    def forward(self, input_image: torch.Tensor) -> pystiche.LossDict:
+        for encoder in self._multi_layer_encoders:
+            encoder.encode(input_image)
+
+        loss = pystiche.LossDict(
+            [(name, op(input_image)) for name, op in self.named_children()]
+        )
+
+        for key in loss.keys():
+            loss[key] = loss[key] / input_image.size()[0]
```
pmeier: What are you doing here?

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,167 @@
...
+            content_loss_kwargs = {}
+        content_loss = ulyanov_et_al_2016_content_loss(
+            impl_params=impl_params,
+            instance_norm=instance_norm,
+            multi_layer_encoder=multi_layer_encoder,
+            **content_loss_kwargs,
+        )
+        return UlyanovEtAl2016PerceptualLoss(
+            style_loss, content_loss=content_loss, stylization=stylization
+        )
```
pmeier: In combination with the above comment, you could simply set `content_loss = None` and pass it to `UlyanovEtAl2016PerceptualLoss`.

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: The authors use the images from the _ImageNet  ILSVRC2012_ and the resolution of these images varies. For this reason they must have preprocessed the images, but there is no information about this step.

pmeier: Ah, OK. In that case I suggest we use `CenterCrop`, since this most of the time captures the motif of the image.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,192 @@
...
+    def get_lr(self) -> List[float]:
+        exp = self.last_epoch - self.delay + 1
+        if exp > 0:
+            return [base_lr * self.gamma ** exp for base_lr in self.base_lrs]
+        else:
+            return self.base_lrs
+
+
+def ulyanov_et_al_2018_lr_scheduler(
+    optimizer: Optional[Optimizer] = None, impl_params: bool = True,
```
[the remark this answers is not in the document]
jbueltemeier: But for the `optimizer` the `transformer` is needed. For this reason, if the `optimizer` is not known here, `none` should be returned and this should be handled later when the `transformer` is known. For example in the `default_transformer_epoch_optim_loop` as it is done there.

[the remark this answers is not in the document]
pmeier: 1. You cannot create an `LRScheduler` without an `optimizer`. Try calling `ulyanov_et_al_2018_lr_scheduler()` without parameters and this will raise an `Exception`.
2. You are right that an `LRScheduler` ultimately requires the `transformer` first, but I don't see how this is a problem. If I'm missing something, could you describe an scenario where this is a problem?
3. Just saw that: it should be `2016` instead of `2018` in the signature.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,192 @@
...
+
+    return transformer
+
+
+def ulyanov_et_al_2016_stylization(
+    input_image: torch.Tensor,
+    transformer: Union[nn.Module, str],
+    impl_params: bool = True,
+    instance_norm: bool = None,
+    sample_size: int = 256,
```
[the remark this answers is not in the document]
jbueltemeier: `sample_size` is a flag to change the stylized image size during evaluation. Since this is only useful in the texture, a separate method `ulyanov_et_al_2016_texture` should be created.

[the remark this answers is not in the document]
pmeier: `UlyanovEtAl2016Transformer` if created with `stylization=False` can handle `image: torch.Tensor` as well as `image_size: Tuple[int, int]` as input. In the former case the `image_size` is simply extracted from the passed `image`. Thus, you could create the new function and simply allow for both inputs, i.e. `input: Union[Tuple[int, int], torch.Tensor]`. As a last remark please name the function `ulyanov_et_al_2016_texture_generation` to better reflect what it is doing.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,192 @@
...
+    postprocessor: ulyanov_et_al_2016_postprocessor = None,
+):
+    device = input_image.device
+    if isinstance(transformer, str):
+        style = transformer
+        transformer = ulyanov_et_al_2016_transformer(
+            style=style, impl_params=impl_params, instance_norm=instance_norm,
+        )
+        transformer = transformer.eval()
```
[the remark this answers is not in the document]
jbueltemeier: He uses `model:evaluate()` only in his master branch. In `texture_nets_v1` he does not set it to evaluation mode before the stylization process or texture sampling. I made it conditional so that this step is only skipped in the branch `texture_nets_v1`.

pystiche/papers/ulyanov_et_al_2016/loss.py
jbueltemeier: The loss is divided here by the batch_size, just like the author did in his programming.

pmeier: The `MSEEncodingOperator` and the `GramOperator` internally use [`mse_loss`](https://github.com/pytorch/pytorch/blob/36e36eff2f8d1f8b695d20c89f27d3f27867e6c5/torch/nn/functional.py#L2516) from `torch.nn.functional`. The default loss `reduction` is `"mean"` and I think this includes averaging over all batches. Could you check that please?

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: I currently use images from the `Caltech256` dataset to test the program. In Paper and Branch `master`  `ImageNet  ILSVRC2012`  data is used, whereas in the master branch only images with a size `>256` for height and width are used. There is no information whether he used the data from the training, the validation or both. In Branch `texture_nets_v1` only the validation images from `ImageNet  ILSVRC2012`  are used for the training.

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: He uses threads for his `dataloader` in the branch `master`, where the default setting is `nthreads = 4`. Is this not the same as in this case `num_workers`?

pystiche/papers/ulyanov_et_al_2016/loss.py
jbueltemeier: Yes, the default loss `reduction` is `"mean"` and this includes averaging over all batches. But the author also uses the default settings of  `nn.MSECriterion` in his implementation and then divides again by the `batch_size` before the error is backpropagated.
[https://github.com/torch/nn/blob/master/doc/criterion.md#nn.MSECriterion ](url)

pystiche/papers/ulyanov_et_al_2016/data.py
pmeier: > I currently use images from the `Caltech256` dataset to test the program.

For your tests this should be OK, but we should try to use the same data as the author in the end. 

I'm a little confused. There are only 4 cases:

- `impl_params==False`, `instance_norm==False`
- `impl_params==False`, `instance_norm==True`
- `impl_params==True`, `instance_norm==False`
- `impl_params==True`, `instance_norm==True`

Could you list all information you can find on the used images for each case. Otherwise its hard to make a decision.

pystiche/papers/ulyanov_et_al_2016/loss.py
pmeier: The link is not working, please fix it. The URL that is visible is different from the linked one. The syntax should be `[text that is shown](url that is linked)`. In your case that could be 

[nn.MSECriterion](https://github.com/torch/nn/blob/master/doc/criterion.md#nn.MSECriterion)

---

In that case I would create custom `Operator`s that implement this functionality in `calculate_score`. Check out the [`GatysEckerBethge2015MSEEncodingOperator`](https://github.com/pmeier/pystiche/blob/f5d42058af399b5ba162c439d7c6c42396315e9f/pystiche/papers/gatys_ecker_bethge_2015/loss.py#L27-L38) for an example how to do that.

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: `impl_params==True`, `instance_norm==True`:
- images extracted at random from the ` ImageNet ILSVRC2012` data
- Check `image_size` for `height > 256` and `width > 256 ` otherwise take a different image

`impl_params==True`, `instance_norm==False`:
- images extracted from the ` ImageNet ILSVRC2012` validation data
- No random order

`impl_params==False`, `instance_norm==False` || `impl_params==False`, `instance_norm==True`:
- images extracted at random from the ` ImageNet ILSVRC2012` data

pmeier: Thanks for the summary. A few further questions. I'll number the cases with 1 to 4:

1. As mentioned somewhere else, I think the straight forward approach is to simply center crop images. 
2. Same as 1.
3. The author performed random cropping, correct? If yes, does he do it at runtime, i.e. a new random crop for every time the image is processed, or once upfront? 
4. Same as 3.

In all cases I suggest we implement the random selection with a set random seed to make our  implementation reproducible. I can handle this once the rest of this PR is finished.

pystiche/papers/ulyanov_et_al_2016/data.py
pmeier: Yes, that should be the same but since this has only an effect on the training speed and not the performance we should stick to the default value which is `0`.

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: 1. Correct, because there is no information about the preprocessing of the images, center crop is now used. 
2. Same as 1.
3. Wrong, in this case the images are resized to the correct size with a `'bilinear'` `scale` and stored in a `hdf5` file for training (once upfront). At the same time the `repr` on `relu4_2` of the images are also stored in this file. 
4. Correct, he uses random cropping during runtime.  So he performs the preprocessing every time a new batch is created.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,192 @@
...
+    style_image = style_transform(style_image)
+    style_image = batch_up_image(style_image, loader=content_image_loader)
+
+    criterion.set_style_image(style_image)
+
+    def criterion_update_fn(input_image, criterion, preprocessor=None):
+        if hasattr(criterion, "content_loss"):
+            if preprocessor is None:
+                preprocessor = ulyanov_et_al_2016_preprocessor()
+            preprocessor = preprocessor.to(device)
```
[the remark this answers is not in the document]
jbueltemeier: Do you mean the functions `default_transformer_epoch_optim_loop`  and the functions associated with it should get one more parameter `preprocessor`?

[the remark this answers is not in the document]
pmeier: Nope. Right now you create `preprocessor` within `criterion_update_fn`. That means that every time `criterion_update_fn` is called, i.e. on every batch, the `preprocessor` is created over and over again. Since the `preprocessor` does not need to change for every batch you can create it on the outside of `criterion_update_fn` and simply use it within.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,200 @@
...
+    transformer: Union[nn.Module, str],
+    impl_params: bool = True,
+    instance_norm: bool = False,
+    stylization: bool = True,
+    postprocessor: Optional[CaffePostprocessing] = None,
+):
+    if isinstance(input, torch.Tensor):
+        device = input.device
+    else:
+        device = "cuda" if torch.cuda.is_available() else "cpu"
```
pmeier: You should wrap this in `torch.device()`

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,200 @@
...
+        content_transform = ulyanov_et_al_2016_content_transform(impl_params=impl_params,
+                                                                 instance_norm=instance_norm)
+        content_transform = content_transform.to(device)
+        input_image = content_transform(input_image)
+
+        if postprocessor is None:
+            postprocessor = ulyanov_et_al_2016_postprocessor()
+        postprocessor = postprocessor.to(device)
+        output_image = transformer(input_image)
+        output_image = torch.clamp(postprocessor(output_image), 0, 1)
```
pmeier: Clamping here is obsolete, since it will be performed in the I/O functions anyway.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,200 @@
...
+        if instance_norm or not impl_params:
+            transformer = transformer.eval()
+        transformer = transformer.to(device)
+
+    with torch.no_grad():
+        if postprocessor is None:
+            postprocessor = ulyanov_et_al_2016_postprocessor()
+        postprocessor = postprocessor.to(device)
+        output_image = transformer(input)
+        output_image = torch.clamp(postprocessor(output_image), 0, 1)
```
pmeier: See above comment about clamping.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+    transforms = []
+    if impl_params:
+        if instance_norm:
+            transforms.append(CenterCrop(edge_size))
+        else:
+            transforms.append(
+                Resize((edge_size, edge_size), interpolation_mode="bilinear")
+            )
+    else:
+        transforms.append(CenterCrop(edge_size))
```
pmeier: Shouldn't be a random crop here somewhere?

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+    instance_norm: bool = True,
+    edge_size: Optional[int] = None,
+) -> ComposedTransform:
+    if edge_size is None:
+        edge_size = 256
+
+    if impl_params:
+        interpolation_mode = "bicubic" if instance_norm else "bilinear"
+    else:
+        interpolation_mode = "bilinear"
```
pmeier: You can shorten this in two cases:

```python
if impl_params and instance_norm:
    ...
else:
    ...

File: pystiche/papers/ulyanov_et_al_2016/data.py [hunk 42]
```diff
@@ -0,0 +1,269 @@
...
+        ),
+        "tiger": DownloadableImage(
+            urljoin(content_base_ulyanov, "tiger.jpg"),
+            md5="e82bf374da425fb2c2e2a35a5a751989",
+        ),
+        "neckarfront": DownloadableImage(
+            "https://upload.wikimedia.org/wikipedia/commons/0/00/Tuebingen_Neckarfront.jpg",
+            title="Tübingen Neckarfront",
+            author="Andreas Praefcke",
+            license=CreativeCommonsLicense(("by",), version="3.0"),
```
pmeier: Shouldn't this be `("by", "sa")`?

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+            md5="dc9ad203263f34352e18bc29b03e1066",
+            file="tuebingen_neckarfront__andreas_praefcke.jpg",
+        ),
+        "CheHigh": DownloadableImage(
```
pmeier: Please format all identifiers with lower case letters and separated with underscores.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+        ),
+        "pebble": DownloadableImage(
+            urljoin(base_ulyanov_suppl_texture, "pebble.png"),
+            md5="5b5e5aa6c579e42e268058a94d683a6c",
+        ),
+        "pixelcity_windows2": DownloadableImage(
+            urljoin(base_ulyanov_suppl_texture, "pixelcity_windows2.jpg"),
+            md5="53026a8411e7c26e959e36d3223f3b8f",
+        ),
+        "red-peppers256.o": DownloadableImage(
```
pmeier: The identifier does not have to be same as the file name but rather should be intuitive. Thus, please remove the unneeded information.

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+    root: str,
+    impl_params: bool = True,
+    instance_norm: bool = True,
+    transform: Optional[Transform] = None,
+):
+    if transform is None:
+        transform = ulyanov_et_al_2016_content_transform(
+            impl_params=impl_params, instance_norm=instance_norm
+        )
+    return ImageFolderDataset(root, transform=transform)
```
pmeier: Couldn't we use `ImageNet` from `torchvision` here? If yes, I will explain how.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,165 @@
...
+from pystiche.loss import MultiOperatorLoss
+from .utils import ulyanov_et_al_2016_multi_layer_encoder
+
+
+class UlyanovEtAl2016MSEEncodingOperator(MSEEncodingOperator):
+    def __init__(
+        self, encoder: Encoder, score_weight: float = 1e0
+    ):
+        super().__init__(encoder, score_weight=score_weight)
+        self.loss_reduction = "mean"
```
pmeier: Since the default `loss_reduction` is `"mean"` there is not reason to set it explicitly.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def calculate_score(self, input_repr, target_repr, ctx):
+        score_correction_factor = 1 / input_repr.size()[0]
```
pmeier: `score_correction_factor` was used because some authors included an actual factor like `1/2` in their loss calculations. Here `batch_size` would be a better name. You can extract it conveniently with `extract_batch_size` from `pystiche.image`.

pmeier: This correction should only happen `if impl_params`, correct? If yes, I would pass `impl_params` to the constructor and set a variable like `self.double_batch_size_mean = impl_params`. Within in `calculate_score` you could than do:

```python
score = super().calculate_score(input_repr, target_repr)
if not self.double_batch_size_mean:
    return score

...
```

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,165 @@
...
+            score_weight = 1e0
+
+    if multi_layer_encoder is None:
+        multi_layer_encoder = ulyanov_et_al_2016_multi_layer_encoder()
+    encoder = multi_layer_encoder[layer]
+
+    return UlyanovEtAl2016MSEEncodingOperator(encoder, score_weight=score_weight)
+
+
+class UlyanovEtAl2016GramOperator(GramOperator):
```
pmeier: See comments of `UlyanovEtAl2016MSEEncodingOperator`

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class UlyanovEtAl2016PerceptualLoss(MultiOperatorLoss):
+    def __init__(
+        self,
+        style_loss: MultiLayerEncodingOperator,
+        content_loss: MSEEncodingOperator = None,
+    ) -> None:
+
+        dict = OrderedDict([("style_loss", style_loss)])
```
pmeier: If possible do not use builtin types such as `dict` as variable name. After that line you can no longer call `dict()` to create a new dictionary. 

Here I would simply create a list `modules` and add the losses to it. I only would construct the `OrderedDict` within the call of the super constructor.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,165 @@
...
+        layer_weights=layer_weights,
+        score_weight=score_weight,
+    )
+
+
+class UlyanovEtAl2016PerceptualLoss(MultiOperatorLoss):
+    def __init__(
+        self,
+        style_loss: MultiLayerEncodingOperator,
+        content_loss: MSEEncodingOperator = None,
```
pmeier: 1. `MSEEncodingOperator` needs to be wrapped in `Optional[]`. Otherwise `None` is not a valid argument.
2. Please remove the default value, since the user needs to explicitly state with the value if he wants content loss or not.

pystiche/papers/ulyanov_et_al_2016/utils.py
```diff
@@ -0,0 +1,54 @@
...
+            return [base_lr * self.gamma ** exp for base_lr in self.base_lrs]
+        else:
+            return self.base_lrs
+
+
+def ulyanov_et_al_2016_lr_scheduler(
+    optimizer: Optimizer, impl_params: bool = True,
+) -> Optional[ExponentialLR]:
+    if impl_params:
+        lr_scheduler = ExponentialLR(optimizer, 0.8)
```
pmeier: To increase clarity you could pass the `gamma` argument in keyword style, i.e. `ExponentialLR(optimizer, gamma=0.8)`

pystiche/papers/ulyanov_et_al_2016/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    else:
+        lr_scheduler = DelayedExponentialLR(optimizer, 0.7, 5)
```
pmeier: See comment above.

pmeier:
On a side note: you seem to have switched email addresses for your commits. The later one is not recognized by GitHub. To add multiple email addresses to your account, visit `Settings -> Emails` by clicking on the downward arrow next to your profile picture in the top right corner.

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: As discussed above, `ImageNet` should be used here. So yes. 
The following things should be possible: 
1. Selection between training data, validation data or both (depending on the parameters `instance_norm` and `impl_params`).
2. A random or fixed sequence of data. 
3. Only images with a certain image size (`height > 256` and `width > 256`)

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: What does this `"sa"` mean in this context? 

If this has to be changed here, then this should also be done in the implementation of `gatys_ecker_bethge_2015`.

pmeier: I was wrong, `"sa"` is not needed here. See https://commons.wikimedia.org/wiki/File:Tuebingen_Neckarfront.jpg

---

`"by"` and `"sa"` refer to the [license conditions](https://creativecommons.org/use-remix/cc-licenses/).

pystiche/papers/ulyanov_et_al_2016/data.py
pmeier: I'll take that up. Until then, just continue with what we have now.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+    style_transform = style_transform.to(device)
+    preprocessor = ulyanov_et_al_2016_preprocessor()
+    preprocessor = preprocessor.to(device)
+    style_image = style_transform(style_image)
+    style_image = preprocessor(style_image)
+    style_image = batch_up_image(style_image, loader=content_image_loader)
+    criterion.set_style_image(style_image)
+
+    def criterion_update_fn(input_image, criterion):
```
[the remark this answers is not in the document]
jbueltemeier: If `stylization` is `false`, nothing is done in the function. Is `pass` the right command? 

```python
    if stylization:
        def criterion_update_fn(input_image, criterion):
            criterion.set_content_image(preprocessor(input_image))
    else:
        def criterion_update_fn(input_image, criterion):
            pass
```

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: Yes, a random crop should be performed for `impl_params and instance_norm`. `RandomCrop` still has to be implemented as a subclass of `Transform`. Probably with a set random seed to make our implementation reproducible?

pystiche/papers/ulyanov_et_al_2016/core.py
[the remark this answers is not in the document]
pmeier: Since in python the indentation is part of the language, you need to provide at least a single indented command after commands like `def`. If nothing should be done inside the block, the `pass` keyword is the correct choice.

pystiche/papers/ulyanov_et_al_2016/data.py
pmeier: `torchvision` already implements a [`RandomCrop` transform](https://github.com/pytorch/vision/blob/3c254fb7af5f8af252c24e89949c54a3461ff0be/torchvision/transforms/transforms.py#L399). Maybe we can use it as a starting point. I'll take that up and get back to you when its done.

jbueltemeier: It would be great if you did.

pmeier: It is done. Pull to get the latest version.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,205 @@
...
+
+    if stylization:
+        def criterion_update_fn(input_image, criterion):
+            criterion.set_content_image(preprocessor(input_image))
+    else:
+        def criterion_update_fn(input_image, criterion):
+            pass
+
+
+    default_transformer_epoch_optim_loop(
```
pmeier: Is there a reason this is split from the `return`? This will work here since `transformer` is adapted in-place, but I suggest we simply return `default_transformer_epoch_optim_loop`.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,205 @@
...
+        lr_scheduler=lr_scheduler,
+        quiet=quiet,
+        logger=logger,
+        log_fn=log_fn,
+    )
+
+    return transformer
+
+
+def ulyanov_et_al_2016_style_generation(
```
pmeier: `style_generation` sounds odd here. Maybe `stylization`?

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,205 @@
...
+    return output_image.detach()
+
+
+def ulyanov_et_al_2016_texture_generation(
+    input: Union[Tuple[int, int], torch.Tensor],
+    transformer: Union[nn.Module, str],
+    impl_params: bool = True,
+    instance_norm: bool = False,
+    stylization: bool = True,
+    postprocessor: Optional[CaffePostprocessing] = None,
```
pmeier: Same as above.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,205 @@
...
+):
+    if isinstance(input, torch.Tensor):
+        device = input.device
+    else:
+        device = torch.device()
```
pmeier: This will not work. You need to call `torch.device` with a `str` of the device you want. See [here](https://pytorch.org/docs/stable/tensor_attributes.html#torch.torch.device). A device-agnostic approach, i.e. should work regardless of what device you are on is this:

```python
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+    instance_norm: bool = True,
+    edge_size: Optional[int] = None,
+) -> ComposedTransform:
+    if edge_size is None:
+        edge_size = 256
+
+    if impl_params and instance_norm:
+        interpolation_mode = "bicubic"
+    else:
+        interpolation_mode = "bilinear"
```
pmeier: We can shorten this even more with

```python
interpolation_mode = "bicubic" if impl_params and instance_norm "bilinear"
```

I think this is still legible. Do you agree?

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+        ),
+        "bricks": DownloadableImage(
+            urljoin(base_ulyanov_suppl_texture, "bricks.png"),
+            md5="1e13818e1fbefbd22f110a1c2f781d40",
+        ),
+        "pebble": DownloadableImage(
+            urljoin(base_ulyanov_suppl_texture, "pebble.png"),
+            md5="5b5e5aa6c579e42e268058a94d683a6c",
+        ),
+        "pixelcity_windows2": DownloadableImage(
```
pmeier: This is still confusing. Does this image portray a city or windows?

pystiche/papers/ulyanov_et_al_2016/data.py
```diff
@@ -0,0 +1,269 @@
...
+            urljoin(base_ulyanov_suppl_texture, "pixelcity_windows2.jpg"),
+            md5="53026a8411e7c26e959e36d3223f3b8f",
+        ),
+        "red_peppers": DownloadableImage(
```
pmeier: I think we can drop the `red`. Do you agree?

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,172 @@
...
+class UlyanovEtAl2016MSEEncodingOperator(MSEEncodingOperator):
+    def __init__(
+        self, encoder: Encoder, score_weight: float = 1e0, impl_params: bool = True,
+    ):
+        super().__init__(encoder, score_weight=score_weight)
+        self.double_batch_size_mean = impl_params
+
+
+    def calculate_score(self, input_repr, target_repr, ctx):
+        score = super().calculate_score(input_repr, target_repr, ctx=ctx)
```
pmeier: Is there a reason you call with a keyword argument `ctx=ctx`? If not please remove.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,172 @@
...
+        if not self.normalize_by_num_channels:
+            return gram_matrix
+
+        num_channels = gram_matrix.size()[-1]
+        return gram_matrix / num_channels
+
+    def calculate_score(
+        self, input_repr: torch.Tensor, target_repr: torch.Tensor, ctx: None
+    ) -> torch.Tensor:
+        score = super().calculate_score(input_repr, target_repr, ctx=ctx)
```
pmeier: Same as above.

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,172 @@
...
+class UlyanovEtAl2016PerceptualLoss(MultiOperatorLoss):
+    def __init__(
+        self,
+        style_loss: MultiLayerEncodingOperator,
+        content_loss: Optional[MSEEncodingOperator],
+    ) -> None:
+
+        modules = [("style_loss", style_loss)]
+        if content_loss is not None:
+            modules += [("content_loss", content_loss)]
```
pmeier: Please use [`modules.append()`](https://www.tutorialspoint.com/python/list_append.htm) since this only adds the new item to the already existing list. With your approach you create a new list with your item and after that another new list that contains all elements. While the execution speed is negligible here, we should do it "the right way".

pystiche/papers/ulyanov_et_al_2016/loss.py
```diff
@@ -0,0 +1,172 @@
...
+    return MultiLayerEncodingOperator(
+        multi_layer_encoder,
+        layers,
+        get_encoding_op,
+        layer_weights=layer_weights,
+        score_weight=score_weight,
+    )
+
+
+class UlyanovEtAl2016PerceptualLoss(MultiOperatorLoss):
```
pmeier: I've added a `PerceptualLoss` base class in #120. I think we can use it here. Please have a look.

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: The image represents pixels of different grey value intensity. 
I have now simply called it [pixel_city](https://raw.githubusercontent.com/DmitryUlyanov/texture_nets/texture_nets_v1/supplementary//texture_models/pixelcity_windows2.jpg).

pystiche/papers/ulyanov_et_al_2016/data.py
jbueltemeier: I agree. I erroneously assumed that the author himself used the term `red_peppers`, but he also uses `peppers` in his paper. I changed it.

pystiche/papers/ulyanov_et_al_2016/data.py
pmeier: If the author does not use a specific term in his paper, I would prefer `pixels`.

pystiche/papers/ulyanov_et_al_2016/loss.py
jbueltemeier: This is a required positional argument in `calculate_score()` in the class `GramOperator`. Without this a `TypeError` is created. However, this is not used there. Should this be removed in the class `GramOperator`?

pmeier: You misunderstood: I want to know why you pass it as keyword argument, i.e. `calculate_score(..., ctx=ctx)` instead as positional `calculate_score(..., ctx)`. Even if we dont use it `GramOperator`, it has to be there to be in compliance with the [base method](https://github.com/pmeier/pystiche/blob/cad5ab6e9485680f2543cf4397d0d21e72a88b9e/pystiche/ops/op.py#L228-L235).

pystiche/papers/ulyanov_et_al_2016/core.py
jbueltemeier: This is because the images are not preprocessed for the Transformer. Neither before they go into the transformer nor after. Therefore the Transformer will transform preprocessed images and this step must be undone. There are better solutions for this but the author does it this way.

pmeier: Just checking if I understand you correctly: the `transformer` needs to learn the preprocessing during training, since it is not included as transform, correct? Thus, if we want to perform a stylization or texture generation with the `transformer` we need to postprocess the generated image, correct?

If the top paragraph is correct, I think we need to do one of two things:

1. We either include the `preprocessor` as an optional argument in `ulyanov_et_al_2016_training` or
2. we remove the optional `postprocessor` from `ulyanov_et_al_2016_style_generation` and `ulyanov_et_al_2016_texture_generation`.

We either give the user the change to explicitly set both or none. If we fix the `preprocessor` it would be nonsensical to leave the `postprocessor` up to the user, since only one will give useful results. Thoughts?

pystiche/papers/ulyanov_et_al_2016/loss.py
pmeier: On a second thought, I don't think it will fit here. I designed it to always have a `content` and `style_loss` and optional `regularization`. Since the texture generation has no `content_loss` I think we should stick to the current implementation.

jbueltemeier: Right I did not see that the `content_loss` here is not optional. However, wouldn't it make sense to design it that way? My thought is that it often happens that in the papers a distinction is made between `texture` and `stylization` and the `PerceptualLoss` in both situations only differ in the `content_loss` and the additional function `set_content_image`. I'm not sure, but the default `PerceptualLoss` for the `texture` will differ as little between papers as it does for `stylization`, correct?

pmeier: Yeah, you are right. I see two ways to implement this:

1. Make the `content_loss` optional in the current implementation or
2. Create an `StyleTransferPerceptualLoss` and `TextureSynthesisPerceptualLoss` for the two different cases.

What do you think?

pystiche/papers/ulyanov_et_al_2016/core.py
jbueltemeier: First your top paragragh is correct. 

What advantages would it have if the user could still explicitly set both? If these do not outweigh, I would prefer variant 2 and remove the optional `postprocessor`, because this is the reprogramming of the paper and the author uses only these processors.

pystiche/papers/ulyanov_et_al_2016/loss.py
jbueltemeier: That's a good question. On the one hand, variant 1 is more flexible but not as intuitive, because you don't see directly that you can switch between `texture` and `stylization` with the optional `content_loss`. On the other hand, variant 2 is more intuitive, because you can see directly what the `PerceptualLoss` is for. However, an additional class/function is required that returns the respective `PerceptualLoss`. If you have no other arguments, I would prefer variant 2 for better readability. Maybe with a `default` class that can handle the switch between `texture` and `stylization` with the flag `stylization`? 

Regarding the optional `style_loss`: Is an `NST` without `style_loss` even possible? I have not seen a paper that doesn't use `style_loss`, which would make the optional inclusion necessary.

pmeier: > Is an NST without `style_loss` even possible?

Nope, but neither is one without `content_loss`. I would classify the different types like this

| type                   | `content_loss` | `style_loss` |
| ---------------------- | -------------- | ------------ |
| texture synthesis      | `False`        | `True`       |
| content reconstruction | `True`         | `False`      |
| NST                    | `True`         | `True`       |

The `regularization` is optional for every type. With that in mind I tend to option 1. If later someone complains about this being unintuitive, we can always build the special classes for every type on top.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,207 @@
...
+
+def ulyanov_et_al_2016_texture_generation(
+    input: Union[Tuple[int, int], torch.Tensor],
+    transformer: Union[nn.Module, str],
+    impl_params: bool = True,
+    instance_norm: bool = False,
+    stylization: bool = True,
+):
+    if isinstance(input, torch.Tensor):
+        device = input.device()
```
pmeier: The attribute `Tensor.device()` is not callable. If you simply want to extract the `device` use `input.device`.

pystiche/papers/ulyanov_et_al_2016/core.py
```diff
@@ -0,0 +1,207 @@
...
+        input_image = content_transform(input_image)
+        postprocessor = ulyanov_et_al_2016_postprocessor()
+        postprocessor = postprocessor.to(device)
+        output_image = transformer(input_image)
+        output_image = postprocessor(output_image)
+
+    return output_image.detach()
+
+
+def ulyanov_et_al_2016_texture_generation(
```
pmeier: Lets rename this to `*_texture_synthesis`.

PR #96, opened 2020-03-12
Add default_transformer_epoch_optim_loop

pmeier:
Based on [comment](https://github.com/pmeier/pystiche/pull/88#discussion_r391459374)

---

This adds a default optimization loop for transformers that need to be trained for more than one epoch. It simply calls `default_transformer_optim_loop()` within each epoch and optionally performs `lr_scheduler.step()` afterwards.

Since an `optimizer` wrapped by an `lr_scheduler` has to stay the same object between the epochs, the optional argument `get_optimizer` of `default_transformer_optim_loop()` is replaced by `optimizer`. This breaks BC. If `optimizer=None` is passed the same behavior as before is retained.

### DIFF
--- a/pystiche/optim/log.py
+++ b/pystiche/optim/log.py
@@ -3,6 +3,8 @@ import contextlib
 import sys
 import logging
 import torch
+from torch.optim.optimizer import Optimizer
+from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
 import pystiche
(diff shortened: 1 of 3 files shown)


pmeier:
With the newest commit `default_transformer_optim_loop()` raises a `UserWarning` if `get_optimizer` is used. Thus, this no longer breaks BC.

PR #144, opened 2020-04-01
Fix ETAMeter if no ETA can be calculated

pmeier:


pmeier:
Ci build for reference: https://travis-ci.org/github/pmeier/pystiche/builds/669573166

PR #158, opened 2020-04-03
Add more tests for pystiche.image

pmeier:


pystiche/image/transforms/core.py
```diff
@@ -39,11 +39,11 @@ def compose_transforms(
     def unroll(
         transform: Union[Transform, ComposedTransform]
     ) -> Iterable[Union[Transform, ComposedTransform]]:
-        if isinstance(transform, Transform):
-            return (transform,)
-        elif isinstance(transform, ComposedTransform):
+        if isinstance(transform, ComposedTransform):
             return transform.children()
+        elif isinstance(transform, Transform):
+            return (transform,)
```
pmeier, line 45: The order has to be changes since `ComposedTransform`s are `Transform`s. Otherwise the `elif` clause would never be reached.

pystiche/image/transforms/crop.py
```diff
@@ -103,7 +103,7 @@ def randint(range: int) -> int:
             return torch.randint(range + 1, ()).item()
 
         vert_origin = randint(image_height - crop_height)
-        horz_origin = randint(image_height - crop_height)
+        horz_origin = randint(image_width - crop_width)
```
pmeier, line 106: This was not detected since the test also used `ValidRandomCrop.get_random_origin()` to create the crop origin.

PR #165, opened 2020-04-08
Add isort and flake8 to format checks

pmeier:


pmeier:
Merging this since the discovered format problems should be addressed in separate PRs.

PR #167, opened 2020-04-08
Fix flake8

pmeier:


.flake8
```diff
@@ -13,7 +13,11 @@ exclude =
     __pycache__,
     *.pyc,
     docs/build,
-    docs/source/auto_*
+    docs/source/auto_*,
+    tutorials # for now
```
pmeier, line 17: This line should be removed in the future. We ignore this for now to avoid rebuilding the tutorials.

PR #233, opened 2020-05-07
Add static type checking with mypy

pmeier:
This supersedes #108.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,5 @@
+[mypy]
+
+allow_redefinition = True
+show_error_codes = True
+pretty = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,5 +1,6 @@
 [mypy]
 
+files = pystiche/core
 allow_redefinition = True
 show_error_codes = True
 pretty = True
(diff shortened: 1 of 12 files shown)


mypy.ini
```diff
@@ -0,0 +1,6 @@
+[mypy]
+
+files = pystiche/core
```
pmeier, line 3: Right now we only check the core package. The remaining packages should be checked in separate PRs.

pystiche/misc/misc.py
```diff
@@ -58,28 +59,33 @@ def prod(iterable: Iterable) -> Any:
...
+def _to_nd_arg(dims: int) -> Callable[[Union[T, Sequence[T]]], Tuple[T, ...]]:
+    def to_nd_arg(x: Union[T, Sequence[T]]) -> Tuple[T, ...]:
+        if x is None:
+            msg = build_deprecation_message(
+                "Passing None as argument",
+                "0.4.0",
+                info="If you need this behavior, please implement it in the caller.",
+            )
+            warnings.warn(msg)
+            return None
```
pmeier, line 74: With the old behavior this code cannot be type checked properly.

PR #252, opened 2020-05-19
add docstrings for essential image I/O functions

pmeier:


pmeier:
`intersphinx` was split off to #253

PR #264, opened 2020-05-26
Add type hints to pystiche.optim

pmeier:


pystiche/optim/optim.py
```diff
@@ -182,10 +184,11 @@ def default_transformer_optim_loop(
 
         loading_time = time.time() - loading_time_start
 
-        def closure():
+        def closure() -> float:
             processing_time_start = time.time()
 
-            optimizer.zero_grad()
+            # See XXX
+            optimizer.zero_grad()  # type: ignore[union-attr]
```
pmeier, line 191: `mypy` is unable to infer the correct type inside the closure (see python/mypy/issues/2608). While `optimizer` can be `None` when calling `default_transformer_optim_loop` at this point it is an `Optimizer` and thus `zero_grad` is a valid attribute. We ignore these errors for now and wait for a fix in `mypy`.

PR #266, opened 2020-05-28
Use installed pystiche info as base for docs

pmeier:


docs/source/conf.py
```diff
@@ -43,7 +43,6 @@ def get_bool_env_var(name, default=False):
     or get_bool_env_var("CI")
 )
 
-
```
pmeier, line 46: This is an artifact.

PR #328, opened 2020-06-19
use pytorch_wheel_installer instead of custom code

pmeier:
This removes all custom code to select and install the correct wheel for CI and uses [`pytorch_wheel_installer`](/pmeier/pytorch_wheel_installer) instead.


pmeier:
Superseded by #336

PR #348, opened 2020-07-28
Bump PyTorch versions

pmeier:
This bump makes it necessary to introduce a `make_torch_ge_1_6_compatible` into the tests that adds missing attributes to the stored `DataLoader`s and `Module`s. I expect similar errors in the future. Thus, we need to overthink the test concept in the (near) future.


pmeier:
Test failures are unrelated and will be fixed in #350

Issue #404, opened 2020-09-24
fix download and storing of the demo images


pmeier:
During his review @soumith encountered MD5 checksum mismatch while trying to run the test suite and the examples (https://github.com/pyOpenSci/software-review/issues/25#issuecomment-698135408). The images linked from Pixabay indeed changed, but without a perceptible visual change. 

The checksums have to be adapted and the download tests should be run on a schedule to faster detect these kind of changes we have no control over. If this happens again, we should consider hosting them ourselves to avoid this in the future.

PR #409, opened 2020-09-25
Run test suite on a schedule

pmeier:
Addresses #404. Tests are now run on a schedule. The download tests are separated since they should be triggered sparsely.


pmeier:
Test failures are expected and will be fixed in a follow-up PR.

PR #410, opened 2020-09-26
robustify download_file

pmeier:
Instead of mindlessly writing the response to a file, we now check the response code and optionally the MD5 checksum if provided.


pmeier:
Test failures are unrelated in will be fixed in #411.

PR #411, opened 2020-09-26
fix checksums of demo images from Pixabay

### LINKED ISSUE #404: fix download and storing of the demo images

pmeier:
Fixes #404.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,14 +109,3 @@ jobs:
 
       - name: Run tests
         run: tox -e tests-galleries
-
-      - name: List pystiche.home()
-        if: failure()
-        run: md5sum "${{ steps.pystiche-home.outputs.dir }}/*"
-
-      - name: Upload demo images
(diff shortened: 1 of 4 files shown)


pmeier:
As suspected in #404, images downloaded from [Pixabay](https://pixabay.com/) might include very subtle changes that are not perceptible to the human observer, but change the checksums. An example are the two images below.

Since the [Pixabay license](https://pixabay.com/service/license/) is very permissive and only forbids redistributing on stock or wallpaper sites, we simply host the images ourselves.

paint

paint

Issue #434, opened 2020-10-27
Simplify logging


pmeier:
Currently logging during optimization is unreasonable complex for a library where the optimization is not the core feature. This is especially evident in the usage examples where `demo.logger()` is used to remove the clutter.

Thus we should simplify this quite a bit. I propose to pull [`tqdm`](https://pypi.org/project/tqdm/) as requirement and let it handle the logging. We can customize it a show parts of the loss. 

This change is BC breaking since it would remove [`pystiche.optim.log`](https://github.com/pmeier/pystiche/blob/ad9b3ea09f9803249b162fbe009db25869e42fb4/pystiche/optim/log.py) and [`pystiche.optim.meter`](https://github.com/pmeier/pystiche/blob/ad9b3ea09f9803249b162fbe009db25869e42fb4/pystiche/optim/meter.py) as they are obsolete if using `tqdm`. Furthermore the parameters 

https://github.com/pmeier/pystiche/blob/ad9b3ea09f9803249b162fbe009db25869e42fb4/pystiche/optim/optim.py#L81-L86

as well as

https://github.com/pmeier/pystiche/blob/ad9b3ea09f9803249b162fbe009db25869e42fb4/pystiche/optim/optim.py#L166-L168

and 

https://github.com/pmeier/pystiche/blob/ad9b3ea09f9803249b162fbe009db25869e42fb4/pystiche/optim/optim.py#L373-L375

will be removed.

PR #442, opened 2020-10-28
emit deprecation warnings for custom logging functions

pmeier:
This foreshadows #434


pmeier:
Lint error is unrelated and was fixed in #443.

PR #462, opened 2020-11-03
Add example for NST via model optimization

pmeier:
Cc @t-vi. We could port the weights from the [PyTorch example](https://github.com/pytorch/examples/tree/master/fast_neural_style), since we are fully compatible. I'm waiting for https://github.com/pytorch/examples/issues/837 if this possible within the license.


t-vi:
Cool, thank you!

PR #472, opened 2020-11-11
Replace pystiche.image.transforms with kornia

### LINKED ISSUE #382: Adapt kornia
The sole purpose of the `pystiche.image.transforms` package was to have GPU capable transforms. Other than that they are out of scope for a package about NST. Thus, we'll replace it with [`kornia`](https://kornia.org) where we need it.

This change will not happen on the `master` but rather on the [`kornia`](https://github.com/pmeier/pystiche/tree/kornia) branch.

pmeier:
Closes #382.


pystiche/image/_transforms.py
```diff
@@ -0,0 +1,129 @@
...
+it is fixed upstream. The following improvements are included:
+
+- The default value of 'align_corners' (resize, Affine) now works by default with every
+  interpolation mode
+- Tensor attributes of Transformations (Affine, Normalize, Denormalize) are now
+  registered as non-persistent buffers by default.
+- Parameters that should be passed as tensors (Affine, Normalize, Denormalize) are now
+  allowed as float or tuple of floats for convenience
+- The Affine transformation now is independent of the batch_size
+"""
```
pmeier, line 13: Unfortunately, as of `kornia==0.4.1` we need a compatibility layer, since `kornia` is not able to handle our use case completely.

PR #488, opened 2021-04-28
unpin pytorch

pmeier:


pmeier:
Failures for galleries tests will be fixed in a follow-up PR.

PR #495, opened 2021-05-30
add CI tests for all supported minor PyTorch releases

pmeier:
As of #488 `pystiche` can run with multiple PyTorch versions. Thus, we should test all of them to make sure we don't include BC incompatible changes.


pmeier:
The change added in #489 is not BC compatible since `nn.Module.register_full_backward_hook` was only introduced in `torch==1.8.0`.

PR #502, opened 2021-07-25
remove kornia once again

pmeier:
In #382 we added support for `kornia`. Since then, `torchvision` actually added a lot transformations that work directly on tensors. Since we only have very basic needs, we can use them over `kornia` and drop one extra dependency,


pmeier:
We need to drop support for `torch==1.6.0` / `torchvision==0.7.0` since `torchvision.transforms.resize` does not support tensors.

PR #506, opened 2021-08-02
change default behavior of patch extraction from non-batched to batched

pmeier:
Follow-up to #446.


pmeier:
Test failure is unrelated

Issue #510, opened 2021-08-08
Fix CI workflows to run on `main` branch


pmeier:
Our workflows should all run for pushes on the default branch, for example if a PR is merged. We recently switched the name of the default branch from `master` to `main`, but the workflows weren't updated

https://github.com/pystiche/pystiche/blob/65f4d787e44b1ffbf7e5b6e48298ed8c7460e5a9/.github/workflows/tests.yml#L6

sreeja97:
Hi,
I am new to the Open source community and this issue is marked as a Good first issue. Can I please work on this?

pmeier:
Hey @sreeja97 and welcome to the OSS community and `pystiche` in general :wave: Yes, it would be awesome if you could take this up. 

Please have a look at our [contributing guide lines](https://github.com/pystiche/pystiche/blob/main/CONTRIBUTING.rst) in case you don't know how to proceed from here. If you hit any blockers, feel free to ping me here. I'm happy to guide you through.

pmeier:
Fixed in #519.

PR #519, opened 2021-08-09
changes to make workflows run on main branch instead of master

sreeja97:
The changes are to make CI checks run on the main branch instead of master in reference to issue number  [510](https://github.com/pystiche/pystiche/issues/510)


pmeier (approved):
Awesome, thanks a lot @sreeja97!

PR #520, opened 2021-08-11
Test suite clean-up: Tests in `core` and `data`

### LINKED ISSUE #509: Test suite cleanup

```python
def test_Foo_bar():
    ...

def test_Foo_baz():
    ...
```

instead of 

```python
class TestFoo:
    def test_bar(self):
        ...

    def test_baz(self):
        ...
```

DevPranjal:
Partially fixes #509 
Note: The diffs automatically detected by git seem to be very messed up.

### DIFF
--- a/tests/integration/core/test_utils.py
+++ b/tests/integration/core/test_utils.py
@@ -8,7 +8,7 @@ import pystiche
 import tests.utils
 
 
-class TestUtils:
+class TestExtractPatches:
     @pytest.mark.parametrize(
         ("extract_fn", "dims"),
         [
@@ -23,7 +23,7 @@ class TestUtils:
             )
         ],
     )
-    def test_extract_patchesnd_num_patches(self, extract_fn, dims):
+    def test_patchesnd_num_patches(self, extract_fn, dims):
         batch_size = 2
         num_channels = 1
         size = 64
@@ -43,7 +43,7 @@ class TestUtils:
             *(patch_size,) * dims,
         )
 
-    def test_extract_patches1d(self):
+    def test_patches1d(self):
         batch_size = 2
         num_channels = 1
         length = 9
--- a/tests/integration/core/test_utils.py
+++ b/tests/integration/core/test_utils.py
@@ -8,61 +8,61 @@ import pystiche
 import tests.utils
 
 
-@pytest.mark.parametrize(
-    ("extract_fn", "dims"),
-    [
-        pytest.param(extract_fn, dims, id=tests.utils.extract_fn_name(extract_fn))
-        for dims, extract_fn in enumerate(
-            (
-                pystiche.extract_patches1d,
-                pystiche.extract_patches2d,
-                pystiche.extract_patches3d,
-            ),
-            1,
+class TestUtils:
+    @pytest.mark.parametrize(
+        ("extract_fn", "dims"),
+        [
+            pytest.param(extract_fn, dims, id=tests.utils.extract_fn_name(extract_fn))
+            for dims, extract_fn in enumerate(
+                (
+                    pystiche.extract_patches1d,
+                    pystiche.extract_patches2d,
+                    pystiche.extract_patches3d,
+                ),
+                1,
+            )
+        ],
+    )
+    def test_extract_patchesnd_num_patches(self, extract_fn, dims):
+        batch_size = 2
+        num_channels = 1
+        size = 64
+        patch_size = 3
+        stride = 2
+
+        num_patches_per_dim = int((size - patch_size) // stride + 1)
+        torch.manual_seed(0)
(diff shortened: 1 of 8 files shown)


DevPranjal:
Hi @pmeier, this is a draft PR for a few tests. Could you please verify if I'm making the changes in the right direction?

tests/integration/core/test_utils.py
```diff
@@ -8,61 +8,61 @@
...
-    batch_size = 2
-    num_channels = 1
-    size = 64
-    patch_size = 3
-    stride = 2
-
-    num_patches_per_dim = int((size - patch_size) // stride + 1)
-    torch.manual_seed(0)
-    x = torch.rand(batch_size, num_channels, *[size] * dims)
+class TestUtils:
```
pmeier, line 11: Same as above. The module already namespaces this into `test_utils.py`.

tests/integration/data/test_collections.py
```diff
@@ -13,246 +13,234 @@
...
-def test_DownloadableImage_download_exist(
-    tmpdir, test_image_url, test_image_file, test_image
-):
-    image = data.DownloadableImage(test_image_url)
-    file = path.join(tmpdir, image.file)
-
-    shutil.copyfile(test_image_file, file)
-
-    with pytest.raises(FileExistsError):
+class DownloadableImage:
```
pmeier, line 16:
```suggestion
class TestDownloadableImage:
```

tests/integration/data/test_datasets.py
```diff
@@ -47,48 +47,47 @@ def test_walkupto_depth(image_folder):
...
-
-    dataset = data.ImageFolderDataset(
-        image_folder, transform=transform, importer=importer
-    )
-
-    actual = dataset[0]
-    desired = path.join(image_folder, "transformed_imported_image0.jpg")
-    assert actual == desired
+class TestImageFolderDataset:
+    def test_image_folder_dataset(self, image_folder):
```
pmeier, line 51: We need a better naming scheme for this case to no repeat information. Maybe `test_core`?

pmeier (changes requested):
Thanks for the PR @DevPranjal! Yes, in general you are moving in the right direction. I have some comments inline, but they should be fairly easy to resolve. 

The PR is already quite big and hard to review. So please address my comments and send a another PR for other files as a follow-up.

pmeier:
Test failure is real:

```
=================================== FAILURES ===================================
______ TestDownloadableImageCollection.test_downloadable_image_collection ______
tests/integration/data/test_collections.py:116: in test_downloadable_image_collection
    collection.download(root=tmpdir)
pystiche/data/collections/download.py:200: in download
    image.download(root=root, overwrite=overwrite)
pystiche/data/collections/download.py:119: in download
    file = path.join(root, file)
/opt/hostedtoolcache/Python/3.6.14/x64/lib/python3.6/posixpath.py:80: in join
    a = os.fspath(a)
E   TypeError: expected str, bytes or os.PathLike object, not TestDownloadableImageCollection
```

Plus, the linting workflow also fails. Please run `tox -f lint` before you commit or install the `pre-commit` hooks so this will be done automatically.

tests/integration/data/test_datasets.py
DevPranjal: There are quite a few places where this situation occurs, shall I change all these occurrences to `test_core` or maybe `test_main`.

tests/integration/data/test_licenses.py
```diff
@@ -1,15 +1,16 @@
...
-        data.NoLicense(),
-        data.PublicDomainLicense(),
-        data.ExpiredCopyrightLicense(1970),
-        data.PixabayLicense(),
-        data.CreativeCommonsLicense(("by", "sa"), "3.0"),
-    )
-    for license in licenses:
-        with subtests.test(license=license):
-            assert isinstance(repr(license), str)
+class TestLicenses:
```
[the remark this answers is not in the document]
DevPranjal: Couldn't think of a class name here. Any suggestions?

tests/integration/data/test_datasets.py
pmeier: I like `test_main` better. And yes, please apply this everywhere.

tests/integration/data/test_licenses.py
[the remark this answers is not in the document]
pmeier: No need for a test class here. Just revert the changes in this file, and we are good to go.

pmeier (approved):
> Note: The diffs automatically detected by git seem to be very messed up.

Yeah, it is quite hard to review this. Given the fact that the coverage stays equal, I thin everything should be good to merge. Thanks a lot @DevPranjal!

Issue #514, opened 2021-08-08
Port tests for `pystiche.loss.functional`


pmeier:
In #500 the file `tests/integration/ops/test_functional.py` was missed to be ported to `tests/integration/loss/test_functional.py`

sakshi-06:
Hi, @pmeier - Thanks for the awesome work on psytiche.

I wanted to work on this and subsequently work on the test suite cleanup. I have tested these changes on my local system, but the packages are broken on my system - so it might take one more day to push the changes as a part of a PR. I wished to ask if I could work on this issue if that's okay with you.

Thanks again!

pmeier:
Hey @sakshi-06 that is awesome news! Please go ahead and work on it!

> I have tested these changes on my local system, but the packages are broken on my system

Can you give me a little more detail so I can help you? What is broken or not working?

PR #521, opened 2021-08-11
Port tests for pystiche.loss.functional 

sakshi-06:
This PR ports `pystiche.ops.functional` to   `pystiche.loss.functional`. Please see #514. 

This is my first PR so in case there are any mistakes please feel free to point them out, I will be looking forward to your review. 

cc: @pmeier


tests/integration/loss/test_functional.py
```diff
@@ -0,0 +1,60 @@
+import pytorch_testing_utils as ptu
+
+import torch
+from torch.nn.functional import mse_loss
+
+import pystiche.ops.functional as F
+
+
+class TestFunctionalLoss:
```
pmeier, line 9: Just seeing this: there is no need for this class here as the module name already implies that these are tests for the functionals. Plus, since it covers all test cases, this test case adds no value.

In order to keep the scope of the PR narrow, please revert this change and do a 1-to-1 port of the old file (minus my comment above). If there is something else here that needs to be addressed, we can do that in follow.up PRs.

tests/integration/loss/test_functional.py
```diff
@@ -60,3 +60,4 @@ def get_checkerboard(size):
     actual = F.total_variation_loss(input)
     desired = 2.0
     assert desired == ptu.approx(actual)
+
```
pmeier, line 63:
```suggestion

```

pmeier:
Thanks again @sakshi-06!

PR #522, opened 2021-08-14
Test suite clean-up: Tests in `demo` and `enc/models`

### LINKED ISSUE #509: Test suite cleanup

DevPranjal:
Partially fixes #509 
Few files added to keep the PR small and reviewable.

### DIFF
--- a/tests/integration/core/test_home.py
+++ b/tests/integration/core/test_home.py
@@ -5,22 +5,22 @@ from os import path
 import pystiche
 
 
-def test_home_default():
-    actual = pystiche.home()
-    desired = path.expanduser(path.join("~", ".cache", "pystiche"))
-    assert actual == desired
-
-
-def test_home_env():
-    tmp_dir = tempfile.mkdtemp()
-    pystiche_home = os.getenv("PYSTICHE_HOME")
-    os.environ["PYSTICHE_HOME"] = tmp_dir
-    try:
+class TestHome:
+    def test_default(self):
         actual = pystiche.home()
-        desired = tmp_dir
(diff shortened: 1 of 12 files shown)


DevPranjal:
Hey @pmeier, I just realized that I branched off my previous branch by mistake. Shall I close this PR and fix the issue?

pmeier:
@DevPranjal 

> Bit of a beginner in git here.

Don't worry, everybody has to start somewhere.

> Shall I close this PR and fix the issue? Or do I rebase to the master into this branch?

There are multiple ways to achieve what we want, so there is no right answer here. IMO the following is the easiest:

Since you included commits that aren't on the `main` branch we can't get rid of them with a merge or a rebase. Thus, we create a new branch from the updated `main` and [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) the commit we want to recycle. You can get the `git` hash either from the GitHub UI or with `git log`.

```shell
$ git checkout main
$ git pull
$ git checkout -b even-more-test-refactor
$ git cherry-pick c04ca09194c3622de01707494959ac0054f79716
$ git push
$ git branch -d more-test-refactor
```

Afterwards we can close this PR and start a new one from the new branch.

DevPranjal:
Thanks for the detailed information. Might not be able to help for 2-3 days due to college work, will finish it as soon as that ends. :)

PR #523, opened 2021-08-20
Cleaned up test suite for tests in demo and enc/models

### LINKED ISSUE #509: Test suite cleanup

DevPranjal:
Partially fixes #509 
Few files added to keep the PR small and reviewable.

### DIFF
--- a/tests/integration/demo/test_demo.py
+++ b/tests/integration/demo/test_demo.py
@@ -5,44 +5,45 @@ import torch
 from pystiche import data, demo
 
 
-def test_images_smoke():
-    images = demo.images()
-    assert isinstance(images, data.DownloadableImageCollection)
+class TestDemoImages:
+    def test_images_smoke(self):
+        images = demo.images()
+        assert isinstance(images, data.DownloadableImageCollection)
 
-    names, _ = zip(*iter(images))
-    assert set(names) == {
-        "bird1",
-        "paint",
-        "bird2",
-        "mosaic",
-        "castle",
-        "church",
-        "cliff",
-    }
+        names, _ = zip(*iter(images))
+        assert set(names) == {
+            "bird1",
+            "paint",
+            "bird2",
+            "mosaic",
+            "castle",
+            "church",
+            "cliff",
+        }
+
+    @pytest.mark.parametrize(
+        ("name", "regions"),
+        [
+            ("castle", ("building", "water", "sky")),
+            ("church", ("building", "sky")),
+            ("cliff", ("water",)),
+        ],
+    )
+    def test_images_guides_smoke(self, name, regions):
+        image = demo.images()[name]
+
+        assert isinstance(image.guides, data.DownloadableImageCollection)
+
+        names, _ = zip(*iter(image.guides))
+        assert set(names).issuperset(set(regions))
 
 
-@pytest.mark.parametrize(
-    ("name", "regions"),
-    [
-        ("castle", ("building", "water", "sky")),
-        ("church", ("building", "sky")),
-        ("cliff", ("water",)),
-    ],
-)
-def test_images_guides_smoke(name, regions):
-    image = demo.images()[name]
+class TestDemoTransformers:
+    def test_transformer(self):
+        torch.manual_seed(0)
+        input_image = torch.rand(1, 3, 256, 256)
+        transformer = demo.transformer()
 
-    assert isinstance(image.guides, data.DownloadableImageCollection)
+        output_image = transformer(input_image)
 
-    names, _ = zip(*iter(image.guides))
-    assert set(names).issuperset(set(regions))
-
-
-def test_transformer():
-    torch.manual_seed(0)
-    input_image = torch.rand(1, 3, 256, 256)
-    transformer = demo.transformer()
-
-    output_image = transformer(input_image)
-
-    assert output_image.shape == input_image.shape
+        assert output_image.shape == input_image.shape
--- a/tests/integration/demo/test_demo.py
+++ b/tests/integration/demo/test_demo.py
@@ -5,8 +5,8 @@ import torch
 from pystiche import data, demo
 
 
-class TestDemoImages:
-    def test_images_smoke(self):
+class TestImages:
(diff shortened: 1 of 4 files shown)


tests/integration/demo/test_demo.py
```diff
@@ -5,44 +5,45 @@
...
-def test_transformer():
-    torch.manual_seed(0)
-    input_image = torch.rand(1, 3, 256, 256)
-    transformer = demo.transformer()
-
-    output_image = transformer(input_image)
-
-    assert output_image.shape == input_image.shape
+class TestDemoImages:
+    def test_images_smoke(self):
```
pmeier, line 9:
```suggestion
    def test_smoke(self):
```

tests/integration/demo/test_demo.py
```diff
@@ -5,44 +5,45 @@
...
-
-def test_transformer():
-    torch.manual_seed(0)
-    input_image = torch.rand(1, 3, 256, 256)
-    transformer = demo.transformer()
-
-    output_image = transformer(input_image)
-
-    assert output_image.shape == input_image.shape
+class TestDemoImages:
```
pmeier, line 8:
```suggestion
class TestImages:
```

tests/integration/demo/test_demo.py
```diff
@@ -5,44 +5,45 @@
...
+
+    @pytest.mark.parametrize(
+        ("name", "regions"),
+        [
+            ("castle", ("building", "water", "sky")),
+            ("church", ("building", "sky")),
+            ("cliff", ("water",)),
+        ],
+    )
+    def test_images_guides_smoke(self, name, regions):
```
pmeier, line 32:
```suggestion
    def test_guides_smoke(self, name, regions):
```

tests/integration/demo/test_demo.py
```diff
@@ -5,44 +5,45 @@
...
+
+class TestDemoTransformers:
+    def test_transformer(self):
+        torch.manual_seed(0)
+        input_image = torch.rand(1, 3, 256, 256)
+        transformer = demo.transformer()
+
+        output_image = transformer(input_image)
+
+        assert output_image.shape == input_image.shape
```
pmeier, line 49: Could you revert this? We don't need a test class for a single test. If there will be more tests in the future, we can always add it later.

tests/integration/enc/models/test_vgg.py
```diff
@@ -30,101 +30,104 @@ def vgg_multi_layer_encoder_loaders(vgg_archs):
...
-    for arch, loader in zip(vgg_archs, vgg_multi_layer_encoder_loaders):
-        with subtests.test(arch=arch):
-            asset = enc_asset_loader(arch)
-
-            multi_layer_encoder = loader(
-                pretrained=True,
-                weights="torch",
-                preprocessing=False,
-                allow_inplace=False,
+class TestVGGPretrained:
```
pmeier, line 33: This class should be about the VGG nets. Right now we only have tests that include the pretrained variant, but that should not be exclusive.

```suggestion
class TestVGG:
```

tests/integration/enc/models/test_vgg.py
```diff
@@ -30,101 +30,104 @@ def vgg_multi_layer_encoder_loaders(vgg_archs):
...
+    @pytest.mark.slow
+    def test_main(self, mocker):
```
pmeier, line 35:
```suggestion
    def test_pretrained(self, mocker):
```

tests/integration/enc/models/test_vgg.py
```diff
@@ -30,101 +30,104 @@ def vgg_multi_layer_encoder_loaders(vgg_archs):
...
+            mocks.make_mock_target(
+                "enc", "models", "vgg", "hub", "load_state_dict_from_url"
+            ),
+            return_value=state_dict,
+        )
+
+        model = vgg(pretrained=True)
+        ptu.assert_allclose(model.state_dict(), state_dict)
+
+    def test_num_classes_mismatch(self):
```
pmeier, line 47:
```suggestion
    def test_pretrained_num_classes_mismatch(self):
```

pmeier (commented):
Thanks @DevPranjal for the PR. I have a few minor naming suggestions inline. Apart from that it seems with this change we are no longer hitting this line:

https://github.com/pystiche/pystiche/blob/b7d354c4baf0404f519cce66593a6cee968a8d33/pystiche/enc/models/utils.py#L140

From quick look, there is nothing in here that should change this. I'll review more thorough later. Feel free to also do so if you want.

tests/integration/enc/models/test_utils.py
```diff
@@ -113,109 +112,108 @@ def collect_modules(self, inplace: bool):
...
-    multi_layer_encoder_state_dict,
-    other_multi_layer_encoder_state_dict,
-    multi_layer_encoder_state_dict_key_map,
-):
-    inv_key_map = {v: k for k, v in multi_layer_encoder_state_dict_key_map.items()}
-    inv_state_dict = OrderedDict(
-        (inv_key_map[name], tensor)
-        for name, tensor in other_multi_layer_encoder_state_dict.items()
-    )
+    def tets_load_state_dict(
```
DevPranjal, line 148: The coverage issue was due to this misspelling

DevPranjal: The coverage looks fine now :)

pmeier:
> The coverage looks fine now :)

Awesome, thanks a lot @DevPranjal!

---

For the future: you can apply suggestions directly from the GitHub UI. You can either click "Commit suggestion" under every comment. This will create a commit for every suggestion. Otherwise you can also add them to a batch and apply them all together. This way you don't need replicate my comments on your local machine somehow.

pmeier (approved):

DevPranjal:
Noted, thanks!

PR #536, opened 2021-08-24
Test suite clean-up: Tests in `loss` and `image`

DevPranjal:
Addresses #509 :)
A few places where I was skeptical about the grouping of tests, I have mentioned them below

### DIFF
--- a/tests/integration/image/test_utils.py
+++ b/tests/integration/image/test_utils.py
@@ -157,59 +157,62 @@ def test_edge_to_image_size_vert_horz():
     assert actual == desired
 
 
-class TestExtract:
-    def test_batch_size(self):
-        batch_size = 3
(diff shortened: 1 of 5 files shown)


tests/integration/image/test_utils.py
```diff
@@ -157,122 +157,118 @@ def test_edge_to_image_size_vert_horz():
     assert actual == desired
 
 
-def test_extract_batch_size():
-    batch_size = 3
+class TestExtract:
```
pmeier, line 160: No need to group here. Although they all start with `extract`, they are not closely related. Please revert this.

pmeier (approved):
Looking good @DevPranjal! I have one minor comment inline. Otherwise it is good to go!

PR #537, opened 2021-08-25
Test suite clean-up: Tests in `enc`, `meta` and `misc`

DevPranjal:
Addresses #509

### DIFF
--- a/tests/integration/misc/test_misc.py
+++ b/tests/integration/misc/test_misc.py
@@ -118,111 +118,114 @@ def test_verify_str_arg():
     assert actual == desired
 
 
-def test_get_input_image_tensor():
-    image = torch.tensor(0.0)
+class TestGetInputImageTensor:
+    def test_main(self):
+        image = torch.tensor(0.0)
 
-    starting_point = image
-    actual = misc.get_input_image(starting_point)
-    desired = image
-    assert actual is not desired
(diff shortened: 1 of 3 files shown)


DevPranjal:
I skipped [this](https://github.com/pystiche/pystiche/blob/main/tests/integration/enc/test_multi_layer_encoder.py) file, but there is a solo function at the end which is out of the namespace. Should I group the functions in this file?

tests/integration/misc/test_misc.py
```diff
@@ -118,111 +118,108 @@ def test_verify_str_arg():
...
-    starting_point = "content"
-    image = torch.tensor(0.0)
-
-    actual = misc.get_input_image(starting_point, content_image=image)
-    desired = image
-    assert actual == ptu.approx(desired)
-
-    with pytest.raises(RuntimeError):
-        misc.get_input_image(starting_point, style_image=image)
+class TestGetInputImageTensor:
```
pmeier, line 121: It seems the tests still have a legacy name. Let's rename them 
```suggestion
class TestGetInputImage:
```

pmeier (commented):
> Should I group the functions in this file?

Yes. As a rule-of-thumb: a class will always get a separate test case even if there is only a single test in there.

PR #548, opened 2021-09-01
`tqdm` usage in `example/beginner/example_nst_without_pystiche.py`

### LINKED ISSUE #546: Use `tqdm` in beginner example without `pystiche`
Since  we use `tqdm` for our logging in `pystiche`, we should also use it in comparison example without `pystiche`:

https://github.com/pystiche/pystiche/blob/4fb1f7e425fd195b35284d7cbf7cd90b84be4333/examples/beginner/example_nst_without_pystiche.py#L446-L470

DevPranjal:
Fixes #546

### DIFF
--- a/examples/beginner/example_nst_without_pystiche.py
+++ b/examples/beginner/example_nst_without_pystiche.py
@@ -39,6 +39,8 @@ from torchvision import transforms
 from torchvision.models import vgg19
 from torchvision.transforms.functional import resize
 
+from tqdm.auto import tqdm
+
 print(f"I'm working with torch=={torch.__version__}")
 print(f"I'm working with torchvision=={torchvision.__version__}")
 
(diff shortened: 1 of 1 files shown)


examples/beginner/example_nst_without_pystiche.py
```diff
@@ -444,30 +445,31 @@ def show_image(image, title=None):
 # . This structure is also valid for all other optimizers.
 
 num_steps = 500
+iterable = range(1, num_steps + 1)
```
pmeier, line 448: Do we need the `iterable` here? 

```suggestion
```

examples/beginner/example_nst_without_pystiche.py
```diff
@@ -444,30 +445,31 @@ def show_image(image, title=None):
...
 
-for step in range(1, num_steps + 1):
+with tqdm(desc="Image optimization", iterable=iterable) as progress_bar:
+    for _ in iterable:
```
pmeier, line 451:
```suggestion
with tqdm(desc="Image optimization", total=num_steps) as progress_bar:
    for _ in range(num_steps):
```

examples/beginner/example_nst_without_pystiche.py
```diff
@@ -444,30 +445,31 @@ def show_image(image, title=None):
...
+            perceptual_loss.backward()
 
-        if step % 50 == 0:
-            print(f"Step {step}")
-            print(f"Content loss: {content_score.item():.3e}")
-            print(f"Style loss:   {style_score.item():.3e}")
-            print("-----------------------")
+            progress_bar.set_postfix(
+                loss=f"{float(perceptual_loss):.3e}", refresh=False
+            )
```
pmeier, line 467: Just to confirm: we are not refreshing here, because we only update afterwards, right?

pmeier (approved):
Thanks a lot for the PR @DevPranjal! One minor suggestion and one question inline. Otherwise LGTM! Since CI will not build the galleries, I'll build them locally later to see if I missed anything in the review.

pmeier (approved):
@DevPranjal I've added the suggested changes. Docs are looking good locally. The change should be up tomorrow :slightly_smiling_face:

examples/beginner/example_nst_without_pystiche.py
DevPranjal: Yes. If I'm not wrong, manual update was necessary since tqdm iterable was not a part of the loop.

DevPranjal:
Thanks for the changes.

PR #566, opened 2021-09-18
Suppress Deprecation Warning

krshrimali:
This PR is in draft mode. Here is what it does so far:

- Implements a decorator (`suppress_depr_warnings`) which captures and ignores `DeprecationWarning`/`UserWarning`.
- Slap the decorator (`suppress_depr_warnings`) on tests present in `tests/integration/loss/*` and `tests/integration/ops/test_comparison.py` files.
- Force `pytest` to fail on warnings (`pytest -Werror`)

Testing with: `tox -e tests-integration` on the local system:

<details>
    <summary>Click to see skips/failures</summary>

```
================================================================================== short test summary info ===================================================================================
SKIPPED [1] tests/integration/test_optim.py:48: Test errors on Python 3.8 only. This is most likely caused by the test itself rather than the code it should test.
SKIPPED [1] tests/integration/test_optim.py:64: Test errors on Python 3.8 only. This is most likely caused by the test itself rather than the code it should test.
SKIPPED [1] tests/integration/test_optim.py:86: Test errors on Python 3.8 only. This is most likely caused by the test itself rather than the code it should test.
SKIPPED [1] tests/integration/test_optim.py:102: Test errors on Python 3.8 only. This is most likely caused by the test itself rather than the code it should test.
SKIPPED [1] tests/integration/test_optim.py:141: Test errors on Python 3.8 only. This is most likely caused by the test itself rather than the code it should test.
SKIPPED [1] tests/integration/test_optim.py:166: Test errors on Python 3.8 only. This is most likely caused by the test itself rather than the code it should test.
SKIPPED [1] tests/integration/core/test_objects.py:223: CUDA is not available.
SKIPPED [1] tests/integration/enc/models/test_alexnet.py:18: Test is flaky and --run-flaky was not given.
SKIPPED [1] tests/integration/enc/models/test_vgg.py:75: Test is flaky and --run-flaky was not given.
FAILED tests/integration/core/test_math.py::TestCosineSimilarity::test_shape - UserWarning: The default value of batched_input has changed from False to True in version 1.0.0. To suppress...
FAILED tests/integration/data/test_collections.py::TestDownloadableImage::test_generate_file - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/data/test_licenses.py::test_licenses_repr_smoke - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/enc/models/test_alexnet.py::TestAlexNetMultiLayerEncoder::test_smoke - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/enc/models/test_alexnet.py::TestAlexNetMultiLayerEncoder::test_state_dict_url - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/enc/models/test_utils.py::TestModelMultiLayerEncoder::test_internal_preprocessing - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/enc/models/test_utils.py::TestModelMultiLayerEncoder::test_repr - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/enc/models/test_vgg.py::TestVGGMultiLayerEncoder::test_smoke - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/enc/models/test_vgg.py::TestVGGMultiLayerEncoder::test_state_dict_url - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/image/test_io.py::test_show_image_smoke - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/loss/test_functional.py::TestMRFLoss::test_main - UserWarning: The default value of batched_input has changed from False to True in version 1.0.0. To suppress thi...
FAILED tests/integration/ops/test_container.py::TestOperatorContainer::test_main - UserWarning: The class pystiche.ops.OperatorContainer is deprecated since pystiche==1.0 and will be remo...
FAILED tests/integration/ops/test_container.py::TestOperatorContainer::test_get_image_or_guide - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/ops/test_container.py::TestOperatorContainer::test_set_image_or_guide - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/ops/test_container.py::TestOperatorContainer::test_call - UserWarning: The class pystiche.ops.OperatorContainer is deprecated since pystiche==1.0 and will be remo...
FAILED tests/integration/ops/test_container.py::TestMultiLayerEncodingOperator::test_main - UserWarning: The class pystiche.ops.MultiLayerEncodingOperator is deprecated since pystiche==1....
FAILED tests/integration/ops/test_container.py::TestMultiRegionOperator::test_set_regional_image_or_guide - pytest.PytestDeprecationWarning: A private pytest class or function was used.
FAILED tests/integration/ops/test_regularization.py::test_TotalVariationOperator_call - UserWarning: The class pystiche.ops.TotalVariationOperator is deprecated since pystiche==1.0 and wi...
FAILED tests/integration/ops/test_regularization.py::test_ValueRangeOperator_call - UserWarning: The class pystiche.ops.ValueRangeOperator is deprecated since pystiche==1.0 and will be re...
FAILED tests/integration/pyramid/test_pyramid.py::TestImagePyramid::test_iter_resize - pytest.PytestDeprecationWarning: A private pytest class or function was used.
========================================================================= 20 failed, 242 passed, 9 skipped in 24.58s =========================================================================
```
</details>

With `tox -e tests-integration`, there are 20 failures - 242 passes - 9 skips _with this branch_ while on the `upstream/main` branch (after adding `-Werror` to `tox.ini`), there were 36 failures - 252 passes - 9 skips.

**TODOs:**

* [ ] Consider replacing the new decorator `suppress_depr_warnings` with existing decorator: `suppress_warnings`.
* [ ] Make sure only those warnings are ignored/suppressed which are coming from `pystiche`. (Since we explicitly use `pystiche` in the deprecation message, I'm thinking of checking that keyword in the message for this.)
* [ ] Is there a better way to slap this decorator on all the tests instead of adding for each test manually?

**Note**:

* This PR is in a heavy WIP mode. But it's also in heavy development mode, should be ready for review in a day.

cc: @pmeier


tox.ini
```diff
@@ -18,7 +18,7 @@ deps =
   igittigitt
   # TODO: move to a released version
   git+https://github.com/pmeier/pytorch_testing_utils
-commands = pytest -c pytest.ini --durations=25
+commands = pytest -Werror -c pytest.ini --durations=25
```
pmeier, line 21: Lets put that into `pytest.ini`, since this should apply to all runs not just the integration tests.

pmeier:
> * Consider replacing the new decorator `suppress_depr_warnings` with existing decorator: `suppress_warnings`.

`suppress_warnings` is not a decorator, but rather a context manager. It is used for the same problem that we are trying to solve, but in our implementation rather than in the test suite. I agree, it would be good to marry the two. Maybe we can have the decorator approach only in `tests.utils`?

```python
from pystiche.misc import (
    suppress_deprecation_warnings as _suppress_deprecation_warnings,
)


def suppress_deprecation_warning(test_fn):
    @functools.wraps(test_fn)
    def wrapper(*args, **kwargs):
        with _suppress_deprecation_warnings():
            return test_fn(*args, **kwargs)

    return wrapper
```

> * Is there a better way to slap this decorator on all the tests instead of adding for each test manually?

Even if there is, I wouldn't go for that. I think it is ok to be verbose here. Especially since we only need to do that once if something is deprecated. Afterwards the test should not change or there shouldn't be any new tests that check deprecated stuff.

tox.ini
krshrimali: Thanks for the suggestion. This should be fixed now in the recent commit.

krshrimali:
> > * Consider replacing the new decorator `suppress_depr_warnings` with existing decorator: `suppress_warnings`.
> 
> `suppress_warnings` is not a decorator, but rather a context manager. It is used for the same problem that we are trying to solve, but in our implementation rather than in the test suite. I agree, it would be good to marry the two. Maybe we can have the decorator approach only in `tests.utils`?
> 
> ```python
> from pystiche.misc import (
>     suppress_deprecation_warnings as _suppress_deprecation_warnings,
> )
> 
> 
> def suppress_deprecation_warning(test_fn):
>     @functools.wraps(test_fn)
>     def wrapper(*args, **kwargs):
>         with _suppress_deprecation_warnings():
>             return test_fn(*args, **kwargs)
> 
>     return wrapper
> ```

I opted with what you suggested since I don't have a strong opinion on this (both look good to me). This should be fixed in the recent commit. I took the freedom to remove `suppress_warnings` which isn't used anywhere in the code, please let me know if this should stay in case you plan to use it in the future.

> 
> > * Is there a better way to slap this decorator on all the tests instead of adding for each test manually?
> 
> Even if there is, I wouldn't go for that. I think it is ok to be verbose here. Especially since we only need to do that once if something is deprecated. Afterwards the test should not change or there shouldn't be any new tests that check deprecated stuff.

Right, for now, I added the decorator on the tests where this warning was raised. Does that sound good? Or should it go for all the tests?

This PR should be ready for review, please take a look whenever you find the time. @pmeier 

Thanks!

pystiche/misc.py
```diff
@@ -280,16 +280,13 @@ def reduce(x: torch.Tensor, reduction: str) -> torch.Tensor:
...
     for category in categories:
-        warnings.filterwarnings("ignore", category=category)
-    new_filters = set(warnings.filters) - old_filters  # type: ignore[attr-defined]
-    try:
-        yield
-    finally:
-        for filter in new_filters:
-            warnings.filters.remove(filter)  # type: ignore[attr-defined]
-        warnings._filters_mutated()  # type: ignore[attr-defined]
+        warnings.filterwarnings(
```
pmeier, line 287: I think just doing `filterwarnings` will set them "forever". The old version had the ability to reset these filters after the context manager is done. I think it would be even better to use [`warnings.catch_warnings`](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) if that is possible.

tests/integration/loss/test_perceptual.py
```diff
@@ -5,8 +5,11 @@
 
 from pystiche import enc, loss, ops
 
+from tests.utils import suppress_deprecation_warning
+
 
 class TestPerceptualLoss:
+    @suppress_deprecation_warning
```
pmeier, line 12: Why did you put the decorator here? IIRC, this should not be deprecated. Same for `set_style_image`.

pystiche/misc.py
```diff
@@ -280,16 +280,13 @@ def reduce(x: torch.Tensor, reduction: str) -> torch.Tensor:
...
+            "ignore",
+            category=category,
+            message=".* is deprecated since pystiche==.* and will be removed in a future release.",
```
pmeier, line 290: I'm a little anxious about matching the message. Can we maybe use the [`module` filter](https://docs.python.org/3/library/warnings.html#describing-warning-filters) to check if a warning originated in `pystiche`?

pmeier (changes requested):
Hey @krshrimali, thanks for the PR! I have a few comments inline. Apart from them two more general ones:

>  I took the freedom to remove `suppress_warnings` which isn't used anywhere in the code, please let me know if this should stay in case you plan to use it in the future.

As of `pystiche==1.0.0` we have adopted semantic versioning. That means, we cannot introduce BC breaking changes on a non-major release (of course we can, but then we wouldn't follow the scheme anymore). The next release should be `1.1.0` so just a minor one. Thus, we cannot remove `suppress_warnings` here. Still we can mark it as deprecated, so it can be removed with the `2.0.0` release.

> Right, for now, I added the decorator on the tests where this warning was raised. Does that sound good? Or should it go for all the tests?

In this PR we should cover all tests where we actually test deprecated stuff. If there are failures in other tests, let's ignore them for now and fix this in follow-up PRs. To do that, I suggest we add the `filterwarnings` option to the `pytest.ini`. [Example](https://github.com/scipy/scipy/blob/aac7ba7cfdde023beada90a5cb8392b527383346/pytest.ini#L5-L14)

krshrimali:
Minor update: Thanks @pmeier for the review. Just a little caught up with PyTorch during the weekdays, will get back on this one with updates by the weekend or earlier. Hope that's fine.

pmeier:
@krshrimali No worries, I'm happy for every contribution!

tests/integration/loss/test_perceptual.py
krshrimali: I think `FeatureReconstructionOperator` is deprecated.

```
FAILED tests/integration/loss/test_perceptual.py::TestPerceptualLoss::test_set_content_image - UserWarning: The class pystiche.ops.FeatureReco...
FAILED tests/integration/loss/test_perceptual.py::TestPerceptualLoss::test_set_style_image - UserWarning: The class pystiche.ops.FeatureRecons...
```

Will have to disable warnings as failures to get the full log for warnings. (will update post dinner)

pystiche/misc.py
krshrimali: How about `module=.*pystiche`? (https://stackoverflow.com/a/53706908) - I've used this in the recent commit, it works well.

pystiche/misc.py
krshrimali: You are right, for some reason - using context manager wasn't working here, though using only decorator works. I've pushed this in the recent commit.

krshrimali:
> Hey @krshrimali, thanks for the PR! I have a few comments inline. Apart from them two more general ones:
> 
> > I took the freedom to remove `suppress_warnings` which isn't used anywhere in the code, please let me know if this should stay in case you plan to use it in the future.
> 
> As of `pystiche==1.0.0` we have adopted semantic versioning. That means, we cannot introduce BC breaking changes on a non-major release (of course we can, but then we wouldn't follow the scheme anymore). The next release should be `1.1.0` so just a minor one. Thus, we cannot remove `suppress_warnings` here. Still we can mark it as deprecated, so it can be removed with the `2.0.0` release.

Sure, I've brought it back to life but will have to mark it deprecated. This should be done in the next commit (after your inputs on existing implementation).

> In this PR we should cover all tests where we actually test deprecated stuff. If there are failures in other tests, let's ignore them for now and fix this in follow-up PRs. To do that, I suggest we add the `filterwarnings` option to the `pytest.ini`. [Example](https://github.com/scipy/scipy/blob/aac7ba7cfdde023beada90a5cb8392b527383346/pytest.ini#L5-L14)

Unfortunately, using `filterwarnings` and `-Werror` together doesn't work. The warnings (as failures) are still raised. I tried with the following `pytest.ini` file:

```python
[pytest]

markers =
  large_download
  slow
  slow_if_cuda_not_available
  flaky

addopts =
  -ra
  -Wignore::UserWarning
  -Wignore::pytest.PytestDeprecationWarning
  -Werror
  --tb=short
```

Also as:

```python
[pytest]

markers =
  large_download
  slow
  slow_if_cuda_not_available
  flaky

filterwarnings=
  ignore::UserWarning
  ignore::pytest.PytestDeprecationWarning

addopts =
  -ra
  -Werror
  --tb=short
```

Do you have any suggestions on this?

Thanks for the patience, @pmeier!

tests/integration/loss/test_perceptual.py
pmeier: > I think `FeatureReconstructionOperator` is deprecated.

Yes, but in such a case we should fix the test. We can simply replace `FeatureReconstructionOperator` with `FeatureReconstructionLoss` and should eliminate the warning.

krshrimali: Alright, sure! I'll fix this.

pystiche/misc.py
```diff
@@ -39,7 +39,7 @@
     "get_device",
     "download_file",
     "reduce",
-    "suppress_warnings",
```
pmeier, line 42: This needs to be reverted.

tests/utils.py
```diff
@@ -183,3 +186,17 @@ def extract_fn_name(fn):
...
+        categories = (UserWarning,)
+    def deprecation_decor(test_fn):
+        @functools.wraps(test_fn)
+        def wrapper(*args, **kwargs):
+            for category in categories:
+                with warnings.catch_warnings():
+                    warnings.filterwarnings("ignore", category=category, module=r'.*pystiche')
+                    return test_fn(*args, **kwargs)
+        return wrapper
+    return deprecation_decor
```
pmeier, line 202: 1. Since we named the function `suppress_deprecation_warning`, IMO it is fair game to not make it configurable to suppress all kind of warnings. Thus, I would remove the outer wrapper.
2. Since you are invoking `test_fn` inside the `categories` loop, you will effectively run the test multiple times and only suppress one warning category each run. We want to register multiple filters based on the `categories`, but only run the test once.
3. The `module` regex is a little to permissive IMO. It allows all modules that have the name `pystiche` in it. I suggest we use `module="pystiche([.].*)?"`. With that `module` can either be `pystiche` or any of its submodules `pystiche.*`

```suggestion
def suppress_deprecation_warning(test_fn):
    @functools.wraps(test_fn)
    def wrapper(*args, **kwargs):
        with warnings.catch_warnings():
            for category in (UserWarning, DeprecationWarning):
                warnings.filterwarnings(
                    "ignore", category=category, module="pystiche([.].*)?"
                )

            return test_fn(*args, **kwargs)

    return wrapper
```

pmeier (commented):
Hey @krshrimali, thanks a lot for the new commits.

> Do you have any suggestions on this?

I meant what you suggested in https://github.com/pystiche/pystiche/pull/566#discussion_r716038547 so we are almost good to go. Only minor suggestions inline.

Apart from them, I see one bigger issue: it seems our new decorator does not play nice with `pytest-subtests`. We might encounter errors that read `E   pytest.PytestDeprecationWarning: A private pytest class or function was used.` I wanted to remove subtests for a while now and replace it with `pytest.mark.parametrize`. We probably need to do that before we merge this PR. Thoughts?

pystiche/misc.py
pmeier: See https://github.com/pystiche/pystiche/pull/566#pullrequestreview-763695045.

krshrimali:
> Hey @krshrimali, thanks a lot for the new commits.
> 
> > Do you have any suggestions on this?
> 
> I meant what you suggested in [#566 (comment)](https://github.com/pystiche/pystiche/pull/566#discussion_r716038547) so we are almost good to go. Only minor suggestions inline.
> 
> Apart from them, I see one bigger issue: it seems our new decorator does not play nice with `pytest-subtests`. We might encounter errors that read `E pytest.PytestDeprecationWarning: A private pytest class or function was used.` I wanted to remove subtests for a while now and replace it with `pytest.mark.parametrize`. We probably need to do that before we merge this PR. Thoughts?

Really good suggestions and questions, @pmeier - acknowledging that I've seen these comments. The suggestion looks good to fix #571, looks like @ankitaS11 is interested so this shouldn't be too late. :) I'll also send in the fixes from your comments tomorrow.

Thank you for taking a look again!

tests/integration/loss/test_perceptual.py
krshrimali: This has been fixed now!

pystiche/misc.py
krshrimali: Done, thanks for pointing this Philip!

PR #567, opened 2021-09-18
add remaining CLI tests

pmeier:


pmeier:
The problem is a memory leak introduced in #565. It was not recognized there, since only with the tests added in this PR we are hitting the upper limit of available memory of the CI.

Issue #571, opened 2021-09-26
remove subtests from test suite


pmeier:
Using the `unittest.TestCase.subTest` context manager was the usual way to run a test against multiple parametrizations or to separate multiple smaller tests. This has multiple disadvantages, from which not being able to run a single subtest is the most severe. Lately, we also discovered, that using a decorator that suppresses certain warnings in tests, also does not work with subtests (#566, cc @krshrimali).

Thus, we should start to remove it. For multiple parametrizations, we can simply replace it with [`pytest.mark.parametrize`](). If we have cases for multiple subtests in one test, we should either "inline" them or break up the test into multiple small ones.

I found references to the `subtest` fixture in the following folders:

- [x] `download` #575
- [x] `download/hub` #575
- [x] `integration` #574
- [x] `integration/data` #572
- [x] `integration/enc/models` #577
- [x] `integration/image` #578
- [x] `integration/loss` #579
- [x] `integration/ops` #580
- [x] `integration/pyramid` #576

Let's go folder by folder :slightly_smiling_face:

ankitaS11:
Hi, @pmeier - got to know about this opened issue from @krshrimali. I was thinking to pick these up, confirming if that's okay?

pmeier:
Hey @ankitaS11, always a pleasure to have new contributors. Welcome! Yes, pick anything from the list and get started. I've linked one that I've. You can use it as inspiration.

Leave the tests directly in the `integration` folder up to me. They are a little tricky.

pmeier:
Hey @ankitaS11 any update on this? I want to move forward with #566 and this needs to be fixed first.

PR #577, opened 2021-10-04
Removed subtests in tests/integration/enc/models

ankitaS11:
This PR addresses issue #571 for `integration/enc/models`.

Note- All tests passed on my local system, please suggest any changes if required.

cc @pmeier

### DIFF
--- a/tests/integration/enc/models/test_alexnet.py
+++ b/tests/integration/enc/models/test_alexnet.py
@@ -8,12 +8,10 @@ from pystiche import enc
 
 
 class TestAlexNetMultiLayerEncoder:
-    def test_smoke(self, subtests):
+    def test_smoke(self):
         multi_layer_encoder = enc.alexnet_multi_layer_encoder(pretrained=False)
         assert isinstance(multi_layer_encoder, enc.alexnet.AlexNetMultiLayerEncoder)
-
-        with subtests.test("repr"):
-            assert isinstance(multi_layer_encoder, enc.alexnet.AlexNetMultiLayerEncoder)
+        assert isinstance(multi_layer_encoder, enc.alexnet.AlexNetMultiLayerEncoder)
 
     @pytest.mark.large_download
     @pytest.mark.slow
@@ -37,21 +35,20 @@ class TestAlexNetMultiLayerEncoder:
         desired = asset.output.enc_keys
         assert actual == desired
 
-    def test_state_dict_url(self, subtests, frameworks):
+    def test_state_dict_url(self, frameworks):
         def should_be_available(framework):
             return framework == "torch"
 
         multi_layer_encoder = enc.alexnet_multi_layer_encoder(pretrained=False)
 
         for framework in frameworks:
-            with subtests.test(framework=framework):
-                if should_be_available(framework):
-                    assert isinstance(
-                        multi_layer_encoder.state_dict_url(framework), str
-                    )
-                else:
-                    with pytest.raises(RuntimeError):
-                        multi_layer_encoder.state_dict_url(framework)
+            if should_be_available(framework):
+                assert isinstance(
+                    multi_layer_encoder.state_dict_url(framework), str
+                )
+            else:
+                with pytest.raises(RuntimeError):
+                    multi_layer_encoder.state_dict_url(framework)
 
     @pytest.mark.slow
     def test_load_state_dict_smoke(self):
--- a/tests/integration/enc/models/test_utils.py
+++ b/tests/integration/enc/models/test_utils.py
@@ -114,18 +114,17 @@ def multi_layer_encoder_cls(
 
 class TestModelMultiLayerEncoder:
     def test_internal_preprocessing(
-        self, subtests, frameworks, multi_layer_encoder_cls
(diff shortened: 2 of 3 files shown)


ankitaS11:
I understand that removing `with subtests.test("repr")` led to 2 tests being flaky:

```bash
SKIPPED [1] tests/integration/enc/models/test_alexnet.py:16: Test is flaky and --run-flaky was not given.
SKIPPED [1] tests/integration/enc/models/test_vgg.py:72: Test is flaky and --run-flaky was not given.
```

I'm not too sure what's the best solution for this, plus I'm not really aware what `subtests.test("repr")` does! Can you please help me with this? @pmeier

pmeier:
> I understand that removing `with subtests.test("repr")` led to 2 tests being flaky:

Nope, they were flaky to begin with:

https://github.com/pystiche/pystiche/blob/218cf56d4c1b070ed7397f7e2218f911523c8a02/tests/integration/enc/models/test_alexnet.py#L20-L21

https://github.com/pystiche/pystiche/blob/218cf56d4c1b070ed7397f7e2218f911523c8a02/tests/integration/enc/models/test_vgg.py#L77-L78

Don't worry about that. 

> I'm not really aware what `subtests.test("repr")` does!

In general, one function constitutes one test. If you check multiple things inside it, we still only get the information if all checks passed or if any failed. Thus, as a rule of thumb, you should only have a single check, e.g. `assert`, per test. However, this is not always practical. Sometimes object creation is verbose or expensive and we don't want to replicate it over multiple tests. In such a situation we might use multiple checks per tests for practicality. 

To improve on this a little one can introduce a subtest. A subtest will simply be registered as individual test inside of another. Still,  there are multiple downsides to this and the goal should be to remove all of its instances.

tests/integration/enc/models/test_alexnet.py
```diff
@@ -8,12 +8,10 @@
...
 
 class TestAlexNetMultiLayerEncoder:
-    def test_smoke(self, subtests):
+    def test_smoke(self):
         multi_layer_encoder = enc.alexnet_multi_layer_encoder(pretrained=False)
         assert isinstance(multi_layer_encoder, enc.alexnet.AlexNetMultiLayerEncoder)
-
-        with subtests.test("repr"):
-            assert isinstance(multi_layer_encoder, enc.alexnet.AlexNetMultiLayerEncoder)
+        assert isinstance(multi_layer_encoder, enc.alexnet.AlexNetMultiLayerEncoder)
```
pmeier, line 14: It would be better to split this into two tests, e.g. `test_main_smoke` and `test_repr_smoke`

pmeier: On that note: the second line is completely redundant :facepalm: It should be something like

```python
assert isinstance(repr(multi_layer_encoder), str)
```

to check if the `repr` can be constructed and is a string.

tests/integration/enc/models/test_alexnet.py
```diff
@@ -37,21 +35,20 @@ def test_main(self, enc_asset_loader):
...
-                else:
-                    with pytest.raises(RuntimeError):
-                        multi_layer_encoder.state_dict_url(framework)
+            if should_be_available(framework):
+                assert isinstance(
+                    multi_layer_encoder.state_dict_url(framework), str
+                )
+            else:
+                with pytest.raises(RuntimeError):
+                    multi_layer_encoder.state_dict_url(framework)
```
pmeier, line 51: By just removing `subtests` here, we lose information. If this test now fails, we can't know which `framework` caused this without manually debugging it. Every time you see `subtests` used in conjunction with a loop, you can probably replace it with a test parametrization. In this case for example:

```python
    @pytest.mark.parametrize(
        ("framework", "should_be_available"),
        [
            pytest.param(framework, should_be_available, id=framework)
            for framework, should_be_available in [("torch", True), ("caffe", False)]
        ],
    )
    def test_state_dict_url(self, framework, should_be_available):
        multi_layer_encoder = enc.alexnet_multi_layer_encoder(pretrained=False)

        if should_be_available:
            assert isinstance(multi_layer_encoder.state_dict_url(framework), str)
        else:
            with pytest.raises(RuntimeError):
                multi_layer_encoder.state_dict_url(framework)
```

What this does, is that the test will actually be instantiated two times. This way we can easily see which parametrization is failing and have successfully removed the `subtests`.

pmeier (changes requested):
Hey @ankitaS11, you are right that after your change we test exactly the same as before, but unfortunately we currently lose some information. I've added a few inline comments to (hopefully) guide you better in the direction I want this to head. Feel free to ask questions in case something is not clear.

ankitaS11:
> Hey @ankitaS11, do you have an update on this? I've finished all the other files, so this is the last one in the stack. Should I take over from here or do you want to finish this?

Hi, @pmeier - I'm sorry, I wasn't doing well for some time and hence the delay. I appreciate your patience, and the ping - I've addressed all the review comments (thanks for the explanation - it was really helpful), and this should be ready for your review. In case you feel there is anything else to address, I'll be happy to make amends without any delay (I'm back to work :)).

Thank you again!

PR #586, opened 2021-11-06
fix mypy for torch==1.10.0

pmeier:


pmeier:
The `_module` dict got some annotations in pytorch/pytorch#61334 which was included in the `torch==1.10.0` release. It seems that `None` was allowed all along for modules, but this is now picked up by `mypy`. Still, unless someone does something stupid, we can be fairly certain that `_modules` will only contain `nn.Modules`. Thus, we use the same approach that PyTorch uses for example for `nn.Sequential` and simply overwrite the annotation to fit our use case.

PR #587, opened 2021-11-07
reenable tdqm patch for gallery builds

pmeier:
Revert accidental change from #583 that disables patching `tqdm` in the galleries


pmeier:
Test failure is unrelated.

PR #591, opened 2021-12-15
Added note in Model Optimisation example pointing out example trainin…

dooglewoogle:
I hadn't touched pytorch or ML much before using pystiche. It took me a day or two to figure out how to load, prepare and conform the images. Naturally I discovered the example script afterwards.

### DIFF
--- a/examples/advanced/example_model_optimization.py
+++ b/examples/advanced/example_model_optimization.py
@@ -279,24 +279,16 @@ show_image(style_image)
 #   If you do not specify an ``optimizer``, the
 #   :func:`~pystiche.optim.default_model_optimizer`, i.e. :class:`~torch.optim.Adam` is
 #   used.
-#
-# .. note::
-#
-#   The weights of the provided transformer were trained with the
-#   `2014 training images <http://images.cocodataset.org/zips/train2014.zip>`_ of the
-#   `COCO dataset <https://cocodataset.org/>`_. The training was performed for
-#   ``num_epochs=2`` and ``batch_size=4``. Each image was center-cropped to
-#   ``256 x 256`` pixels.
 
 
-def train(transformer, root, batch_size=4, epochs=2, image_size=256):
+def train(*, transformer, root, batch_size, epochs, image_size):
     if root is None:
         raise RuntimeError("You forgot to define a root image directory.")
 
-    image_transforms = nn.Sequential(
-        transforms.Resize(image_size), transforms.CenterCrop(image_size)
+    transform = nn.Sequential(
+        transforms.Resize(image_size), transforms.CenterCrop(image_size),
     )
-    dataset = ImageFolderDataset(root, transform=image_transforms)
+    dataset = ImageFolderDataset(root, transform=transform)
 
     from torch.utils.data import DataLoader
 
@@ -312,22 +304,13 @@ def train(transformer, root, batch_size=4, epochs=2, image_size=256):
 ########################################################################################
 # Depending on the dataset and your setup the training can take a couple of hours. To
 # avoid this, we provide transformer weights that were trained with the scheme above.
-#
-# .. note::
-#
-#   If you want to perform the training yourself, set ``root`` to the location
-#   of the images downloaded from the link below
-#
-
-root = None
-checkpoint = "example_transformer.pth"
 
 
-def load(checkpoint):
+def download_example_transformer_weights():
+    checkpoint = "example_transformer.pth"
     if path.exists(checkpoint):
         return torch.load(checkpoint)
-
-    if checkpoint == "example_transformer.pth":
+    else:
 
         @contextlib.contextmanager
         def suppress_output():
@@ -343,23 +326,47 @@ def load(checkpoint):
             state_dict = hub.load_state_dict_from_url(url)
         return state_dict
 
-    else:
-        raise RuntimeError("Please specify valid path to checkpoint")
 
-
-if root:
-    transformer = train(transformer, root)
-    state_dict = OrderedDict(
-        [
-            (name, parameter.detach().cpu())
-            for name, parameter in transformer.state_dict().items()
-        ]
-    )
-    torch.save(state_dict, checkpoint)
-
-else:
-    state_dict = load(checkpoint)
+use_example_transformer_weights = True
+if use_example_transformer_weights:
+    state_dict = download_example_transformer_weights()
     transformer.load_state_dict(state_dict)
+
+########################################################################################
+#
+#
+# .. note::
+#
+#   The weights of the provided transformer were trained with the
+#   `2014 training images <http://images.cocodataset.org/zips/train2014.zip>`_ of the
+#   `COCO dataset <https://cocodataset.org/>`_. The training was performed for
+#   ``num_epochs=2`` and ``batch_size=4``. Each image was center-cropped to
+#   ``256 x 256`` pixels.
+#
+# .. note::
+#
+#   If you want to perform the training yourself, set ``root`` to a location
+#   of a folder of images.
+
+
+root = None
+checkpoint = "example_transformer.pth"
+if not use_example_transformer_weights:
+    if root:
+        transformer = train(
+            transformer=transformer, root=root, batch_size=4, epochs=2, image_size=256
+        )
+        state_dict = OrderedDict(
+            [
+                (name, parameter.detach().cpu())
+                for name, parameter in transformer.state_dict().items()
+            ]
+        )
+        torch.save(state_dict, checkpoint)
+
+    else:
+        state_dict = torch.load(checkpoint)
+        transformer.load_state_dict(state_dict)
 
 
 ########################################################################################
--- a/examples/advanced/example_model_optimization.py
+++ b/examples/advanced/example_model_optimization.py
@@ -22,9 +22,11 @@ from os import path
 import torch
 from torch import hub, nn
 from torch.nn.functional import interpolate
+from torchvision import transforms
 
 import pystiche
 from pystiche import demo, enc, loss, optim
+from pystiche.data import ImageFolderDataset
 from pystiche.image import show_image
 from pystiche.misc import get_device
 
@@ -277,16 +279,24 @@ show_image(style_image)
 #   If you do not specify an ``optimizer``, the
 #   :func:`~pystiche.optim.default_model_optimizer`, i.e. :class:`~torch.optim.Adam` is
 #   used.
+#
+# .. note::
+#
+#   The weights of the provided transformer were trained with the
+#   `2014 training images <http://images.cocodataset.org/zips/train2014.zip>`_ of the
+#   `COCO dataset <https://cocodataset.org/>`_. The training was performed for
+#   ``num_epochs=2`` and ``batch_size=4``. Each image was center-cropped to
+#   ``256 x 256`` pixels.
 
 
-def train(
-    transformer, dataset, batch_size=4, epochs=2,
-):
-    if dataset is None:
-        raise RuntimeError(
-            "You forgot to define a dataset. For example, "
-            "you can use any image dataset from torchvision.datasets."
-        )
+def train(transformer, root, batch_size=4, epochs=2, image_size=256):
+    if root is None:
+        raise RuntimeError("You forgot to define a root image directory.")
+
+    image_transforms = nn.Sequential(
+        transforms.Resize(image_size), transforms.CenterCrop(image_size)
+    )
+    dataset = ImageFolderDataset(root, transform=image_transforms)
 
     from torch.utils.data import DataLoader
 
@@ -305,31 +315,20 @@ def train(
 #
 # .. note::
 #
-#   If you want to perform the training yourself, set
-#   ``use_pretrained_transformer=False``. If you do, you also need to replace
-#   ``dataset = None`` below with the dataset you want to train on.
+#   If you want to perform the training yourself, set ``root`` to the location
+#   of the images downloaded from the link below
 #
-# .. note::
-#
-#   The weights of the provided transformer were trained with the
-#   `2014 training images <http://images.cocodataset.org/zips/train2014.zip>`_ of the
-#   `COCO dataset <https://cocodataset.org/>`_. The training was performed for
-#   ``num_epochs=2`` and ``batch_size=4``. Each image was center-cropped to
-#   ``256 x 256`` pixels.
-#
-#   An example script to load and conform the dataset, then perform the training can be found in the `scripts folder <https://github.com/pystiche/pystiche/blob/main/scripts/perform_model_optimization.py>`_
-#   of the repository
 
-use_pretrained_transformer = True
+root = None
 checkpoint = "example_transformer.pth"
 
-if use_pretrained_transformer:
+
+def load(checkpoint):
     if path.exists(checkpoint):
-        state_dict = torch.load(checkpoint)
-    else:
-        # Unfortunately, torch.hub.load_state_dict_from_url has no option to disable
-        # printing the downloading process. Since this would clutter the output, we
-        # suppress it completely.
+        return torch.load(checkpoint)
+
+    if checkpoint == "example_transformer.pth":
+
         @contextlib.contextmanager
         def suppress_output():
             with open(os.devnull, "w") as devnull:
@@ -342,12 +341,14 @@ if use_pretrained_transformer:
 
         with suppress_output():
             state_dict = hub.load_state_dict_from_url(url)
+        return state_dict
 
-    transformer.load_state_dict(state_dict)
-else:
-    dataset = None
-    transformer = train(transformer, dataset)
+    else:
+        raise RuntimeError("Please specify valid path to checkpoint")
 
+
+if root:
+    transformer = train(transformer, root)
     state_dict = OrderedDict(
         [
             (name, parameter.detach().cpu())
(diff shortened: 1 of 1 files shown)


pmeier:
Worse, `pystiche.transforms` was removed with `pystiche==1.0.0` and thus the snippet is no longer functional. It should be something like:

```python
from torchvision import datasets, transforms

dataset = datasets.ImageFolder(
    root,
    transform=transforms.Compose(
        [
            transforms.Resize(256),
            transforms.CenterCrop(256),
        ]
    ),
)
```

Let me try and run with this and see if we get comparable results.

pmeier:
@dooglewoogle I propose the following solution: Currently we have 2 code blocks. One that defines the training and one for the dispatch. We could split it into three:

1. This block should again contain the training, but in addition the definition for the dataset. Basically, the `train` function should take `root` instead of `dataset`.
2. This block should contain all the logic to load the checkpoint either from local or remote.
3. This block contains the dispatch logic. Something like this:

    ```python
    root = None
    checkpoint = "example_transformer.pth"
    
    if root:
        state_dict = train(transformer, root)
        torch.save(state_dict, checkpoint)
    else:
        state_dict = load(checkpoint)
        transformer.load_state_dict(state_dict)
    ```

With this, a user can simply override `root` with any folder containing images to train. Otherwise the script will simply load pretrained weights. Thoughts?

pmeier:
Hey @dooglewoogle the he issue you raised with this PR is valid to me. Are you planning on going through with the PR or should I take over?

dooglewoogle:
Hi @pmeyer,

Yes, I'd like to do this. I've been on holiday for the last week and will
be until well past new year. I'm planning on picking it up when I get back
in to work, as long as this timeframe is ok with you.

Thanks,
Doug.

pmeier:
Hey Doug, I didn't mean to rush you. Just wanted to confirm that is still on the agenda. Early next year sounds good to me. Let me know if you need help.

pmeier:
Hey @dooglewoogle, just dropping in to say that I didn't forget about this. I was quite busy the last few days and hope to review this in the next couple of days.

examples/advanced/example_model_optimization.py
```diff
@@ -277,16 +279,24 @@ def enc_to_repr(self, enc: torch.Tensor) -> torch.Tensor:
 #   If you do not specify an ``optimizer``, the
 #   :func:`~pystiche.optim.default_model_optimizer`, i.e. :class:`~torch.optim.Adam` is
 #   used.
+#
+# .. note::
```
pmeier, line 283: We should keep this note above the block that actually calls the training since at this point the pretrained weights are not yet introduced.

examples/advanced/example_model_optimization.py
```diff
@@ -277,16 +279,24 @@ def enc_to_repr(self, enc: torch.Tensor) -> torch.Tensor:
...
 
-def train(
-    transformer, dataset, batch_size=4, epochs=2,
-):
-    if dataset is None:
-        raise RuntimeError(
-            "You forgot to define a dataset. For example, "
-            "you can use any image dataset from torchvision.datasets."
-        )
+def train(transformer, root, batch_size=4, epochs=2, image_size=256):
```
pmeier, line 292: Maybe we can remove the defaults and the values to the call? That would link this call to the note above.

```suggestion
def train(*, transformer, root, batch_size, epochs, image_size):
```

examples/advanced/example_model_optimization.py
```diff
@@ -277,16 +279,24 @@ def enc_to_repr(self, enc: torch.Tensor) -> torch.Tensor:
...
+    if root is None:
+        raise RuntimeError("You forgot to define a root image directory.")
+
+    image_transforms = nn.Sequential(
+        transforms.Resize(image_size), transforms.CenterCrop(image_size)
```
pmeier, line 297: Nit: by adding a comma there, `black` puts the transforms on two lines, which IMO makes it easier to read.

```suggestion
        transforms.Resize(image_size), transforms.CenterCrop(image_size),
```

examples/advanced/example_model_optimization.py
```diff
@@ -277,16 +279,24 @@ def enc_to_repr(self, enc: torch.Tensor) -> torch.Tensor:
...
+    )
+    dataset = ImageFolderDataset(root, transform=image_transforms)
```
pmeier, line 299: Let's just keep it as `transform`:

```suggestion
    transform = nn.Sequential(
        transforms.Resize(image_size), transforms.CenterCrop(image_size)
    )
    dataset = ImageFolderDataset(root, transform=transform)
```

examples/advanced/example_model_optimization.py
```diff
@@ -305,28 +315,20 @@ def train(
 #
 # .. note::
 #
-#   If you want to perform the training yourself, set
-#   ``use_pretrained_transformer=False``. If you do, you also need to replace
-#   ``dataset = None`` below with the dataset you want to train on.
+#   If you want to perform the training yourself, set ``root`` to the location
+#   of the images downloaded from the link below
```
pmeier, line 319: There is no link for image download.

```suggestion
#   If you want to perform the training yourself, set ``root`` to a location
#   of an folder of images.
```

examples/advanced/example_model_optimization.py
```diff
@@ -305,28 +315,20 @@ def train(
...
+def load(checkpoint):
     if path.exists(checkpoint):
-        state_dict = torch.load(checkpoint)
-    else:
-        # Unfortunately, torch.hub.load_state_dict_from_url has no option to disable
-        # printing the downloading process. Since this would clutter the output, we
-        # suppress it completely.
+        return torch.load(checkpoint)
+
+    if checkpoint == "example_transformer.pth":
```
pmeier, line 330: What is the use case for this extra check?

examples/advanced/example_model_optimization.py
```diff
@@ -339,12 +341,14 @@ def suppress_output():
...
+
+    else:
+        raise RuntimeError("Please specify valid path to checkpoint")
 
-    transformer.load_state_dict(state_dict)
-else:
-    dataset = None
-    transformer = train(transformer, dataset)
 
+if root:
```
pmeier, line 350: Could we split this block here, by moving the note from the block above as well as the definitions of `root` and `checkpoint` down here?

```suggestion


########################################################################################
# .. note::
#
#   If you want to perform the training yourself, set ``root`` to the location
#   of the images downloaded from the link below

if root:
```

pmeier (commented):
Hey @dooglewoogle, sorry again for the delay. I'm quite happy with the changes. I have a few comments inline about moving the parts around to make the story of the example a little more consistent. Let me know if a comment was not clear enough so I can try to clarify.

examples/advanced/example_model_optimization.py
dooglewoogle: 🤦‍♂️ Changing that

examples/advanced/example_model_optimization.py
dooglewoogle: Its about how best to handle having to download the example transformer. 
If we reach this point we know that the specified transformer does not exist on disk and is the example, so it'll grab it.
We could make this functionality more explicit with a separate download function and a note that this is how to get it?

examples/advanced/example_model_optimization.py
dooglewoogle: Sure, thats fine. Is it a style conform thing?

We would also have to modify the training call later to include them:
`transformer = train(transformer, root, batch_size=4, epochs=2, image_size=256)`

dooglewoogle (commented):
Hi, 
Most of the comments are fine and I've put in. 
Let me know if any of the comments don't make sense, or you'd like more info or discussion. 
Thanks.

examples/advanced/example_model_optimization.py
pmeier: > Sure, thats fine. Is it a style conform thing?

The note that explains the default parameters for training was moved below this block. From a narrative perspective it would thus be off to have them here as default values.



> We would also have to modify the training call later to include them:
> `transformer = train(transformer, root, batch_size=4, epochs=2, image_size=256)`

Yes, we need to modify, but the call should look like:

```py
train(transformer=transformer, root=root, batch_size=4, epochs=2, image_size=256)
```

Due to the `*` in the beginning of the signature, all parameters are keyword only.

examples/advanced/example_model_optimization.py
pmeier: > If we reach this point we know that the specified transformer does not exist on disk and is the example, so it'll grab it.

Maybe I should have been more explicit. Why are we only doing that when the checkpoint matches the default name? Why not always download in case the checkpoint is not available?

> We could make this functionality more explicit with a separate download function and a note that this is how to get it?

Makes sense. Go for it!

dooglewoogle:
I'm not super keen on having added the `use_example_transformer_weights` flag, but needed a reasonably clear way to toggle the example functionality. Please let me know if you have any suggestions about this.

examples/advanced/example_model_optimization.py
```diff
@@ -302,31 +304,14 @@ def train(
...
-#   `2014 training images <http://images.cocodataset.org/zips/train2014.zip>`_ of the
-#   `COCO dataset <https://cocodataset.org/>`_. The training was performed for
-#   ``num_epochs=2`` and ``batch_size=4``. Each image was center-cropped to
-#   ``256 x 256`` pixels.
 
-use_pretrained_transformer = True
-checkpoint = "example_transformer.pth"
 
-if use_pretrained_transformer:
+def download_example_transformer_weights():
```
pmeier, line 309: Let's keep this simple

```suggestion
def download():
```

examples/advanced/example_model_optimization.py
```diff
@@ -302,31 +304,14 @@ def train(
...
+    checkpoint = "example_transformer.pth"
     if path.exists(checkpoint):
-        state_dict = torch.load(checkpoint)
+        return torch.load(checkpoint)
     else:
-        # Unfortunately, torch.hub.load_state_dict_from_url has no option to disable
-        # printing the downloading process. Since this would clutter the output, we
-        # suppress it completely.
+
```
pmeier, line 314: There is no need for an `else` here since `if` branch `return`'s. Please remove the `else` and "unindent" contents by one level.

examples/advanced/example_model_optimization.py
```diff
@@ -302,31 +304,14 @@ def train(
...
-checkpoint = "example_transformer.pth"
 
-if use_pretrained_transformer:
+def download_example_transformer_weights():
+    checkpoint = "example_transformer.pth"
     if path.exists(checkpoint):
-        state_dict = torch.load(checkpoint)
+        return torch.load(checkpoint)
     else:
-        # Unfortunately, torch.hub.load_state_dict_from_url has no option to disable
```
pmeier, line 327: Please put this comment back.

examples/advanced/example_model_optimization.py
```diff
@@ -339,19 +324,49 @@ def suppress_output():
...
-        [
-            (name, parameter.detach().cpu())
-            for name, parameter in transformer.state_dict().items()
-        ]
-    )
-    torch.save(state_dict, checkpoint)
+
+########################################################################################
+#
+#
```
pmeier, line 337:
```suggestion
```

examples/advanced/example_model_optimization.py
```diff
@@ -339,19 +324,49 @@ def suppress_output():
 
         with suppress_output():
             state_dict = hub.load_state_dict_from_url(url)
+        return state_dict
+
 
+use_example_transformer_weights = True
+if use_example_transformer_weights:
+    state_dict = download_example_transformer_weights()
     transformer.load_state_dict(state_dict)
```
pmeier, line 333: Shouldn't this be merged into the next block?

examples/advanced/example_model_optimization.py
```diff
@@ -339,19 +324,49 @@ def suppress_output():
...
+#   If you want to perform the training yourself, set ``root`` to a location
+#   of a folder of images.
+
+
+root = None
+checkpoint = "example_transformer.pth"
+if not use_example_transformer_weights:
+    if root:
+        transformer = train(
+            transformer=transformer, root=root, batch_size=4, epochs=2, image_size=256
```
pmeier, line 357: Nit: put a comma there so the auto-formatter (`tox -e lint-style`) splits this into multiple lines and enhances readability

```suggestion
            transformer=transformer, root=root, batch_size=4, epochs=2, image_size=256,
```

pmeier (commented):
Hey @dooglewoogle sorry for the delay and for being so picky. It seems I'm missing some use case that you are trying to improve. In particular, you haven't answer my question from https://github.com/pystiche/pystiche/pull/591#discussion_r802329280. Let's handle this first before we have another review cycle.

dooglewoogle:
Please don't worry about being nit picky. Honestly, it's not at all a problem. I would rather you end up with code in your project that you are happy with.

I think you are right, we are not quite understanding each other here. The original use case that I started on was loading and transforming/conforming the dataset as this was the only part of the example I felt was not explicitly shown and struggled with. This has been sorted with the `root` and `train()` additions. As far as I'm aware we are just tidying up the rest of the script to make it readable and have a cohesive narrative. 

In regards to [your comment](https://github.com/pystiche/pystiche/pull/591#discussion_r802329280), I see the states the script getting into as:
- Using example transformer, which exists on disk.
- Using example transformer, which does not exist on disk.
- Using custom transformer, which exists on disk.
- Using custom transformer, which does not exist on disk.

The responses to which, would be (in order):
- Load it.
- Download it.
- Load it. 
- Throw Error.

Seeing as there was two states where we just load from disk, what I was attempting [here](https://github.com/dooglewoogle/pystiche/blob/f2f9bc3a77884b5c42536cf2abc695a400665f79/examples/advanced/example_model_optimization.py#L326-L362) was combining those two, then dealing with using the example. Broadly it was:

- Does checkpoint exist on disk?
- - load it, return
- Otherwise:
- - Is example? 
- - - Download it, return
- - Is not example? 
- - - Error

We ended up straying off this for the last series of commits, with the `use_example_transformer_weights` flag, a fileExists check in the download function, and no explicit fileExists check on the custom transformer, leaving that to `torch.load()`

Does this make sense to you? How do you see the states that the script could get into, and how to handle them?

pmeier:
> Using custom transformer, which does not exist on disk.

This is the use case I was missing. I'm guessing you hit this and this is why you are trying to fix it, correct? If yes, could you explain the steps you took to hit this error?

To give some background, this example was originally just meant as that: an example. I did not plan for it to be used as an actual training script that one runs multiple times. That being said, I'm a poor judge of how this stuff is used given that I personally just write minimal scripts from scratch since I know the API well.

dooglewoogle:
> This is the use case I was missing. I'm guessing you hit this and this is why you are trying to fix it, correct? If yes, could you explain the steps you took to hit this error?
> 

I've been digging into this, trying to layout and how I got there, and am reaching the conclusion that I've misunderstood some things too. 
Running the script again as it is, I didn't hit this issue. I think after a couple of ~500MB model downloads I started preferring having them right there in the workspace rather then a cache somewhere else. Also, I habitually check for the presence of files that scripts use. My day job usually has me gluing together multiple other programs and controlling the files they create. Thus the use case of "file does not exist". 
The behaviour of "get the checkpoint, either from disk or url" seems fine to me, and I'm happy to move on from it.


> To give some background, this example was originally just meant as that: an example. I did not plan for it to be used as an actual training script that one runs multiple times. That being said, I'm a poor judge of how this stuff is used given that I personally just write minimal scripts from scratch since I know the API well.

Funny you say that. As a layperson I copied the script and ran it, then started tweaking parameters and images, and ran it again. 
I understood that it was an example, but it was the only example I had that I could start poking further into how transformer setup and training worked.

pmeier:
> I think after a couple of ~500MB model downloads I started preferring having them right there in the workspace rather then a cache somewhere else.

That is a fair complaint. Maybe we can cache it in the default location in `~/.cache/torch/hub/checkpoints`? I think PyTorch has functionality for this so you don't have to download twice. That being said, let's do that in a follow-up PR.

> As a layperson I copied the script and ran it, then started tweaking parameters and images, and ran it again.
I understood that it was an example, but it was the only example I had that I could start poking further into how transformer setup and training worked.

I think this is a fine use case. Still, having a `checkpoint` variable should be sufficient here, no?

---

To bring this PR to an end, I think we can achieve everything with just `root` and `checkpoint`:

```py
if root is None:
    if path.exists(checkpoint):
        # load
    else:
        # download
else:
     # train
```

What do you think?

dooglewoogle:
> That is a fair complaint. Maybe we can cache it in the default location in `~/.cache/torch/hub/checkpoints`? I think PyTorch has functionality for this so you don't have to download twice. That being said, let's do that in a follow-up PR.

This is already implemented, probably as a default from PyTorch

> To bring this PR to an end, I think we can achieve everything with just `root` and `checkpoint`:
> 
> ```python
> if root is None:
>     if path.exists(checkpoint):
>         # load
>     else:
>         # download
> else:
>      # train
> ```
> 
> What do you think?

Sounds great to me. I'll get that in.

pmeier (approved):
Changes look awesome, thanks so much @dooglewoogle! Galleries are only built once per day, so you will only see the changes tomorrow.

PR #593, opened 2021-12-19
fix model optimization training script

pmeier:
Addresses https://github.com/pystiche/pystiche/pull/591#issuecomment-995792201.


pmeier:
Lint failures are unrelated and will be fixed in #594.

Issue #602, opened 2023-03-20
Resurrect?


pmeier:
Unfortunately, I've had no time to properly work on this the past two years. Still, there has been quite some development after the 1.0 release that should not go to waste by never releasing a new version. Thus, I'm committed to at least bring `pystiche` to a state where 1.1 can be released. 

I don't know yet what will come after that. I had hoped for some community engagement to get this project going, but unfortunately (at least for `pystiche`) the field is moving so fast that NST has mostly become obsolete by now. Thus, no matter what features will be released, I don't expect an increase in users in the future. If you read this and feel different, please speak up. 

I'll use this as tracker for the stuff that needs to be done. I'll fight the bit rot first and address bugs or the like afterwards.

pmeier:
From the top of my head:

- [ ] remove `tox` in favor of `doit`
- [ ] Fix gallery build in docs
- [ ] Only support Python >= 3.8
- [ ] #547
- [ ] #598 
- [ ] #599
- [ ] #524
- [x] Move `downloads.pystiche.org` to AWS S3

pmeier:
`torchvision` removed the `model_urls` attribute from their models starting with `torchvision==0.15`

https://github.com/pystiche/pystiche/blob/71217c24557dfba05da5795547bf6f3034e7c66f/pystiche/enc/models/alexnet.py#L5

There is a new alternative starting with `torchvision==0.13` that we should adopt. That would push our minimum required versions

https://github.com/pystiche/pystiche/blob/71217c24557dfba05da5795547bf6f3034e7c66f/setup.cfg#L40-L41

to

```
torch>=1.12
torchvision>=0.13
```

pmeier:
I'm using the [`resurrect` branch](https://github.com/pystiche/pystiche/tree/resurrect) for this.

PR #603, opened 2023-04-11
pin PyTorch dependencies and drop Python 3.6

pmeier:
Temporary fix for https://github.com/pystiche/pystiche/issues/602#issuecomment-1503224470


pmeier:
Ok, bit rot is worse than I thought. I think we need to drop `tox` first to get anything moving here.

